PR #12, opened 2012-03-28
fix clearing problems with hokuyo and costmap

mgruhler:
it is just a hack and meant to be used only temporarily, until either hokuyo driver or costmap_2d has (hopefully) been updated to fix this problem


floweisshardt:
Is there a bugfix we could send to Willow?

I would rather wait for a real fix instead of a work around. Otherwise work arounds typically last for ever...

mgruhler:
The problem seems to be with the hokuyo actually. The docu says, when a max_range measurement occurs, it sends max_range. But it seems that this is not the case and it sends 0.0 instead.

Any hints on how to fix this are welcome :-)

floweisshardt:
for a work around see ticket http://care-o-bot.org/trac/ticket/161

PR #26, opened 2012-10-19
hardware monitor added

uhr-eh:


cob_hwboard/manifest.xml
```diff
@@ -0,0 +1,17 @@
+<package>
+  <description brief="cob_hwboard">
+
+     cob_hwboard
+
+  </description>
+  <author>Eduard Herkel</author>
+  <license>BSD</license>
```
floweisshardt: Please use LGPL

cob_hwboard/manifest.xml
```diff
@@ -0,0 +1,17 @@
...
+  <review status="unreviewed" notes=""/>
+  <url>http://ros.org/wiki/cob_hwboard</url>
```
floweisshardt, line 10: ROS wiki page is still missing, please add some basic documentation there. The header will be filled out as soon as the pull request is in ipa320.

cob_hwboard/src/hwboard.py
```diff
@@ -0,0 +1,705 @@
+#!/usr/bin/env python
```
floweisshardt: please copy and fill ipa header from https://github.com/ipa320/setup/blob/master/templates/license_header.py

cob_hwboard/src/hwboard.py
```diff
@@ -0,0 +1,705 @@
...
+from serial import *
+from diagnostic_msgs.msg import DiagnosticArray
+from diagnostic_msgs.msg import DiagnosticStatus
+from diagnostic_msgs.msg import KeyValue
+
+############################################
+#	check yaml parameters
+############################################
+
+if rospy.has_param("/hwboard/hwboarddevicestring") and rospy.has_param("/hwboard/head_sensor") and rospy.has_param("/hwboard/eye_sensor") and rospy.has_param("/hwboard/torso_module_sensor") and rospy.has_param("/hwboard/torso_sensor") and rospy.has_param("/hwboard/pc_sensor") and rospy.has_param("/hwboard/engine_sensor"):
```
floweisshardt: please use parameters without namespace. The namespace only is added in the launch file, see https://github.com/ipa320/cob_robots/blob/master/cob_bringup/components/lwa.launch

cob_hwboard/src/hwboard.py
```diff
@@ -0,0 +1,705 @@
...
+			retry = retry + 1
+			if preamble_bytes == 7:
+				preamble_bytes = 2
+
+
+############################################
+#	send reset timer message
+############################################
+
+# initialize message and local variables
```
floweisshardt: if all this code is commented, please delete it. If we need it lateron, we have the git history.

cob_hwboard/src/hwboard.py
```diff
@@ -0,0 +1,705 @@
...
+			
+
+				
+#######################################
+#
+#######################################
+
+if __name__ == '__main__':
+	try:
+		hwboard()
```
floweisshardt: why do we need this? an exception should not appear and if it appears we should care about it and not just pass.

PR #28, opened 2012-11-13
fix in cob_light

benmaidel:
No more crashes, if ROBOT environment variable is not set.
If not set, default care o bot light behavior is used. (no inversion of the rgb values)

Note:
Unclear why ROBOT env is not set, when a node is lauched from a different Machine (PC1 is launching node on PC2)


floweisshardt:
I'd rather appreciate a real fix without using the ROBOT environment variable directly in code.

PR #32, opened 2012-12-19
Correct header path of gtkmm.h

frederikhegger:
The header is located under /usr/include/gtkmm-2.4/gtkmm.h (at least under 10.04 and 11.10)

So instead of #include <gtkmm.h> it should be #include <gtkmm-2.4/gtkmm.h>

This fixes the not compiling cob_lightmode package


floweisshardt:
this shouldn't be neccessary. 

#include gtkmm.h should work on all distributions regardless the gtkmm verion. Try with make clean and then make again.

PR #35, opened 2013-01-15
brics oodl driver for sick S300, which can publish scans at 12Hz

janpaulus:
By using a dedicated thread for the communication with the sick s300 I was able to get scans at 12Hz out of it. The publish frequency can be configurte by a parameter.


floweisshardt:
thanks for this pull request, sounds interessting. I'd like to test this intensively before on the hardware, so might take some time to accept this pull request.

see http://www.care-o-bot-research.org/trac/ticket/543

PR #39, opened 2013-01-30
new cob_light driver

benmaidel:
implemented new cob_light driver.

changes:
-priority control (Mode with highest priority will be executed)
-action, service and topic
-different operation modes (Flash, Breath, ColorFade...) can be extended
-special simulation mode (advertises debug topic "debug")

new driver is backward compatible (topic)


cob_light/srv/SetLightMode.srv
```diff
@@ -0,0 +1,4 @@
+cob_light/LightMode mode
+---
+int32 error_type
```
floweisshardt: why does the service have a different return message than the action?

cob_light/msg/LightMode.msg
```diff
@@ -0,0 +1,13 @@
...
+uint8 NONE = 0 # this mode is used if the actual mode sould be used and only the color or frequency should be adapted
+uint8 STATIC = 1 # will change the LEDs to "color"
+uint8 FLASH = 2 # will change the LEDs frequently with "frequency" from "color" black
+uint8 BREATH = 3 # will change the LEDs smoothly with "frequency" from "color" to black
+uint8 BREATH_COLOR = 4 #will change the LEDs smoothly with "frequency" from "color" to black and flips color in time
+uint8 FADE_COLOR = 5 #will fade the colors in rainbow
+
+std_msgs/ColorRGBA color #the color which will be used
+float32 frequency # in Hz
+int32 timeout # in ms
```
floweisshardt: can you explain inside the message definition where we use timeout for. Is there a default value if nothing is specified?

cob_light/msg/LightMode.msg
```diff
@@ -0,0 +1,13 @@
...
+int32 pulses # number of pulses
```
floweisshardt: we should also explain how the pulses value are used. --> maybe it is better to explain all modes and parameter combination on a wiki page.

cob_light/srv/SetLightMode.srv
benmaidel: The intent was to provide an error state for services. Changed this to same return type as action! Return value of the service callback is now used to indicate success or failure.

PR #41, opened 2013-02-28
cob_base_velocity_smoother updates

fmirus:
- the cob_base_velocity_smoother has now a loop rate (adjustable via parameter)
- another new parameter specifies the time interval to stop the robot in case no velocity commands are published
- some bugfixes and code-clean-ups regarding the ros coding rules


floweisshardt:
Can you please update the configuration for all robots (cob+raw+desire) in cob_robots/cob_hardware_config and send a pull request for that too?

PR #50, opened 2013-04-04
cob_phidgets

ipa-josh:


floweisshardt:
please commit configuration (_.yaml) to cob_robots/cob_hardware_config and deployment configuration (_.launch) to cob_robots/cob_bringup --> this shouldn't be in cob_driver

PR #78, opened 2013-08-08
Base struc

cpc-pk:
Parametrierung von Fahrdrehmodulen / ElmoDrives über yaml-Files möglich, keine ini-Files mehr notwendig. 
TODOs:
- Test
- ini-Files "abtippen" in neue yaml Struktur (Undercarriage und Kopfachse)
- Kopfachse-Parametrierung von ini zu yaml umstellen

### DIFF
--- a/cob_base_drive_chain/CMakeLists.txt
+++ b/cob_base_drive_chain/CMakeLists.txt
@@ -32,12 +32,15 @@ rosbuild_gensrv()
 # add include search paths
 INCLUDE_DIRECTORIES(${PROJECT_SOURCE_DIR}/common/include)
 
+# add include search paths
+INCLUDE_DIRECTORIES(${PROJECT_SOURCE_DIR}/ros/include)
+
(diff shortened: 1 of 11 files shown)


mgruhler:
Please do not merge.
This requires a lot more work and testing and is anyways based on an outdated state.

Will be included/considered once a redesing of the respective packages is made.
So please keep the request open until then

floweisshardt:
is supperceeded by new ros_canopen based driver

mgruhler:
@ipa-fmw Fine for me, helps cleaning up the PRs. However, please keep in mind that we still have to consider this depending on how the decision/development regarding r@w's, cob3-x will be driven forward.

So either we need to do something similar (based on the new adapted undercarriage_ctrl which is based on the cob_omni_drive_controller) for the base_drive_chain if we further support it, or we can get rid of this if we support r@w's (required) and probably cob3-x (optional?/required?)

PR #88, opened 2013-10-02
refactored Sick S300 driver for a generic data handling

ipa-josh:
- no fixed point size is assumed anymore
- flexible baud rate
- tested on raw3-1


floweisshardt:
visual review: OK
still to be tested on cob3-3

mgruhler:
@ipa-josh @ipa-fmw 
This should probably not have been merged... There are still some open issues that I discussed with @ipa-josh a while ago (but I cannot remember them right now... ;-) )

PR #89, opened 2013-10-03
Fix setting of baud rate

reed-adept:
Removed forced 500000 baud rate, changed baud rate to be any integer value rather than switching on a small set.


floweisshardt:
outdated

reed-adept:
Is it possible to set the baud rate now with the most recent version of the `cob_sick_s300` driver?

floweisshardt:
@reed-adept we couldn't integrate your PR because it's affecting a lot of packages not relevant to setting the baudrate. Could you send us a clean PR with only modifications in cob_sick_s300?
@ipa-josh is it possible to set the baudrate with the most recent version?

PR #105, opened 2014-03-13
Fixed cob_phidget startup crash

benmaidel:
cob_phidget crashed if there was no configuration on the param server for an attached board


floweisshardt:
I will remove launch and yaml file. This should be in cob_robots or (for testing) be documented on the ROS wiki page

PR #107, opened 2014-03-14
remove launch and yaml configuration. this should be either in cob_robot...

floweisshardt:
...s or on the ROS wiki page


floweisshardt:
visual review: OK

PR #108, opened 2014-03-14
fixed cob_voltage_control groovy build

benmaidel:
phidget21.h does not have a header guard. So this file should be included only one time.


floweisshardt:
visual review: OK

PR #137, opened 2014-05-15
Feature/voltagectrl newphidget

benmaidel:
Changed cob_voltage_ctrl to work with the new phidget driver. (Old version opend phidget directly, new version subscribes to topic with phidget infos)


floweisshardt:
visual review: OK

Issue #134, opened 2014-05-05
separate message files out into standalone message packages


floweisshardt:
e.g. messages in cob_sound (currently cob_script_server depends on cob_driver only due to messages in cob_sound)

mgruhler:
when you are on it, maybe also but the launch files in cob_robots?
- cob_sound
- cob_voltage_control
- cob_cam3d_throttle

fmessmer:
I guess all these issues could be tackled when re-structuring cob_hardware_conifg wrt to separation of X_driver.yaml and X_controller.yaml.

Then also:
- move cob_driver-launch files to cob_robots
- move cob_driver-yaml files to cob_robots
- separate message definitions
  Shall we have a separate repo for messages, cob_srvs, brics_actuator?

floweisshardt:
messages
./cob_relayboard/msg/EmergencyStopState.msg
./cob_phidgets/msg/DigitalSensor.msg
./cob_phidgets/msg/AnalogSensor.msg
./cob_light/msg/LightMode.msg

services
./cob_trajectory_controller/srv/SetFloat.srv
./cob_footprint_observer/srv/GetFootprint.srv
./cob_camera_sensors/srv/GetTOFImages.srv
./cob_camera_sensors/srv/AcquireCalibrationImages.srv
./cob_sound/srv/SayText.srv
./cob_base_drive_chain/srv/ElmoRecorderConfig.srv
./cob_base_drive_chain/srv/ElmoRecorderReadout.srv
./cob_phidgets/srv/SetDigitalSensor.srv
./cob_phidgets/srv/SetDataRate.srv
./cob_phidgets/srv/SetTriggerValue.srv
./cob_light/srv/SetLightMode.srv

actions
./cob_sound/action/Say.action
./cob_light/action/SetLightMode.action

floweisshardt:
done

PR #141, opened 2014-05-21
Bring in velocity/Cartesian controllers

fmessmer:
Thought about bringing in this PR so that we can discuss it in order to improve it.
It's not perfect yet and it still has some known issues:
- still contains yaml's and launch-files (see #134)
- still prone to singularities (cob_twist_controller, cob_lookat_controller)
- handling joints and velocity limits needs to be improved

It has been tested with:
- cob4-1 (simulation) for teleop and odometry_monitor
- raw3-3 (hardware) handle tracking

It is currently not used by default in any robot, only by those using overlays of cob_robots.


floweisshardt:
shouldn't we put that directly into https://github.com/ipa320/cob_control?

fmessmer:
Yes, it will finally go there!
However, I thought, moving all the packages we discussed at once from ipa320 would be the way to go...

floweisshardt:
visual review: OK
opened issue: https://github.com/ipa320/cob_driver/issues/142

PR #147, opened 2014-06-05
Merged with felix and updates for indigo

abubeck:


abubeck:
@ipa-fxm can you approve this?

floweisshardt:
@abubeck you removed the dependency to pr2_msgs and created a new PowerState message (this is good). But in the source code I can only see that you removed the publishers for pr2_msgs/PowerState but without substitution with the new message. Which component is now publishing the new message?

cob_voltage_control/common/src/cob_voltage_control_common.cpp
```diff
@@ -1,7 +1,5 @@
 //ROS typedefs
 #include "ros/ros.h"
-#include <pr2_msgs/PowerBoardState.h>
```
floweisshardt, line 3: please use cob_msgs

cob_voltage_control/ros/src/cob_voltage_control_ros.cpp
```diff
@@ -2,8 +2,6 @@
 #include <ros/ros.h>
 
 // ROS message includes
-#include <pr2_msgs/PowerBoardState.h>
```
floweisshardt, line 5: please use cob_msgs

fmessmer:
PR #151 supersedes this PR and also re-adds the PowerBoardState publisher...
Thus, this PR should be empty once #151 is accepted

PR #148, opened 2014-06-17
Feature/ucar ctrl watchdog

fmirus:
added a watchdog in cob_undercarriage_control to limit the velocity under a threshold to avoid the "racing robot bug"


abubeck:
LGTM

mgruhler:
Is this really not done in the undercarriage_control? How could this have ever worked ;-)

cob_undercarriage_ctrl/ros/src/cob_undercarriage_ctrl.cpp
```diff
@@ -222,13 +244,42 @@ class NodeClass
 		{
 			double vx_cmd_mms, vy_cmd_mms, w_cmd_rads;
 
-			iwatchdog_ = 0;			
+			if(msg->linear.x > max_vel_trans_)
```
mgruhler: Shouldn't you check the absolute value?
It seems to me, you are only checking for positive values.

Holds as well for the other checks below (lines 259, 271)

cob_undercarriage_ctrl/ros/src/cob_undercarriage_ctrl.cpp
```diff
@@ -222,13 +244,42 @@ class NodeClass
...
+			{
+				ROS_DEBUG_STREAM("Recevied cmdVelX: " << msg->linear.x << 
+					", which is bigger than the maximal allowed translational velocity: " <<	max_vel_trans_ << " so set cmdVelX to 0.0");
+				vx_cmd_mms = 0.0;
```
mgruhler: Why do you set the velocity to 0 and not clamp it to the maximum?

Holds as well for the others (l. 263, 275)

fmirus: this is a matter of taste. after some discussions last week we decided to set the velocity to 0 for safety reasons and to show the user that something went wrong with the velocity. But clamping to the maximum is just as applicable.

floweisshardt:
I vote for clamping to the maximum

abubeck:
I vote for clamping to zero. This is the last watchdog before the hardware. If a node really commands a too high velocity, something really bad is happening and we should stop.

fmirus:
:+1:  for clamping to zero. atm we look at each direction separately. if we go for clamping to zero we should also consider to stop the robot completely if undercarriage control receives a really high command in at least one direction. discussions welcome :blush:

abubeck:
We could also think of working with "assert" in this watchdog. This would completely kill the undercarriage_ctrl if we start "drifting" :)

mgruhler:
for the sake of safety: clamp to zero.
But we should then decide on good maximum values (current default is 0.5 m/s). What is the highest command we send from other nodes? I have 0.6 in mind, but I'm not sure.

Depending on that this isn't happening every minute: +1 for an assert

floweisshardt:
for cob4 we've found out that 1.0m/s seems to be the limit. But we're not sure where this limit is defined (ucar, base_drive_chain, elmos)

--> this means we should go for a limit which is definitively higher than 1.0m/s.

mgruhler:
-100 for assert: I just learned the hard way that it gets removed by the preprocessor when compiling in release mode, due to the -NDEBUG flag... (see http://www.cplusplus.com/reference/cassert/assert/)

mgruhler:
oh, and if we receive a command in any direction which is to high, we should definitely stop ALL motion.

floweisshardt:
visual review: OK

mgruhler:
Visual review: OK

fmirus:
any suggestion for reasonable default values for max_velocity?

mgruhler:
I'd say let's take 1.0 m/s.
This is what we use on our robots, if at all. And as it is a parameter it can easily be changed, if anybody wants to drive faster.

What about rotation? 90°/s --> turns around in 2 seconds? I don't really have a fealing what is good here...

floweisshardt:
as we want to drive 1,0m/s I'd like to have at least 1.1m/s as a limit (+10%) before shutting down the system.
90°/s (=1.5708rad/s) in rotation seems reasonable. Again, I'd like to take +10% which would end up in ~1.8rad/sec

mgruhler:
fine for me

mgruhler:
Visual review: OK

floweisshardt:
tested on raw3-1

floweisshardt:
visual review: Ok

PR #152, opened 2014-07-17
Commit of Automatica2014 SMErobotics reveting demo version

lorenz-halt:
Fix: Inversion of sign of digIn
Add: Service to read digital_ios
Fix: Not resulting in error if state to be set is already set


lorenz-halt:
Something is wrong

floweisshardt:
just look like some tabs vs. spaces issue

PR #154, opened 2014-07-26
Can peak sys usb extension

destogl:
Since cob_utilities package is [deprecated](https://raw.githubusercontent.com/iirob/cob_driver/hydro_dev/cob_utilities/package.xml) and in the future "IniFile.h"-files will be removed there is extension of can_generic package to enable set parameters in constructor.

the extension is fully implemented and tested for PEAK-Sys USB dongle, for other CAN devices empty functions are written since I don't have this hardware and can't test them.

### DIFF
--- a/cob_generic_can/common/include/cob_generic_can/CanESD.h
+++ b/cob_generic_can/common/include/cob_generic_can/CanESD.h
@@ -92,6 +92,7 @@ private:
 public:
 	CanESD(const char* cIniFile, bool bObjectMode = false);
 	~CanESD();
+	bool init_ret();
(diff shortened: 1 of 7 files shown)


destogl:
Don't merge yet, I seams there is some error so I can't initialize base any more. Sorry for this, I try to correct it as soon as possible.

destogl:
I'll do it again and add another pull request...

Just a question, should this node read from ROS Paramter Server, or parameters sholud be given trough constructor?

destogl:
Just for info: in a branch is working version.

PR #159, opened 2014-08-25
Proposal for extension of generic_can

destogl:
This proposal is because Ini-config files are deprecated. In this proposal there is not importat which middle-ware is used since the configuration parameters are given during object creation. In my case is used inside ROS, but think is better when doesn't depend on any ROS-specific code (like ROS-parameters).

Tested and used is only PCAN-USB dongle, I don't have possibiliy to test for other CAN interfaces, so empty functions are in other CAN implementations.

Also some constants are added to CanItf.h file to make configuring easier.

**It is tested under indigo, but should also work under hydro.**


floweisshardt:
@ipa-mig: please review

floweisshardt:
@destogl it is ok to merge hydro_dev into indigo_dev but we shouldn't merge indigo_dev back to hydro and especially not the indigo_release_candidate back to hydro_dev because version numbers changed. can you please revert changes so that no merge from indigo_\* branches are in there and resend the PR?

_or_ send PR to indigo_dev.

destogl:
@ipa-fmw Ok, I will make new PR to indigo_dev.

PR #167, opened 2014-08-29
Undo changes on cob_sick_s300 and added record voltage and current scripts

NadiaHG:
Tested on cob4-2: OK


cob_relayboard/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ find_package(catkin REQUIRED COMPONENTS roscpp rostest std_msgs pr2_msgs cob_uti
 
 add_message_files(
   DIRECTORY msg
-  FILES EmergencyStopState.msg
+  FILES EmergencyStopState.msg CurrentMeasurement.msg
```
floweisshardt: please use std_msgs/Float32 (http://docs.ros.org/api/std_msgs/html/msg/Float32.html) instead of an own message.

cob_relayboard/msg/CurrentMeasurement.msg
```diff
@@ -0,0 +1,2 @@
+float32 current_measurement	# value of the current
```
floweisshardt: plese remove this message, see above

cob_voltage_control/ros/src/cob_voltage_control_ros.cpp
```diff
@@ -5,6 +5,7 @@
 #include <pr2_msgs/PowerBoardState.h>
 #include <pr2_msgs/PowerState.h>
 #include <cob_relayboard/EmergencyStopState.h>
+#include <cob_relayboard/CurrentMeasurement.h>
```
floweisshardt: please use std_msgs/Float32 (http://docs.ros.org/api/std_msgs/html/msg/Float32.html) instead of an own message.

floweisshardt:
visual review: OK

PR #169, opened 2014-09-01
Generic can extension

destogl:
This proposal exists because Ini-config files are deprecated. In this proposal there is not importat which middle-ware is used since the configuration parameters are given in constructor. In my case is used inside ROS, but think is better when doesn't depend on any ROS-specific code (like ROS-parameters).

Tested and used is only PCAN-USB dongle, I don't have possibiliy to test for other CAN interfaces, so empty functions are in other CAN implementations.

Also some constants are added to CanItf.h file to make configuring easier.

_Note:_ It would be good to discuss this changes, because than this can be used as a standard-way for slowly transition form Ini-config fils to ROS-Parameter-files. Other possibility is that Can implementations depend on ROS-Parameters explicitly (as now depends on Ini-files), which I don't prefer, because even the ROS is kind a standard in robotics today this code can be used outside ROS, which can have advantages for embedded systems.


floweisshardt:
@ipa-mig: please review. Is same as https://github.com/ipa320/cob_driver/pull/159

cob_generic_can/common/include/cob_generic_can/CanItf.h
```diff
@@ -58,6 +58,21 @@
 #include <cob_generic_can/CanMsg.h>
 //-----------------------------------------------
 
+// for types and baudrates see: https://github.com/ipa320/cob_robots/blob/hydro_dev/cob_hardware_config/raw3-5/config/base/CanCtrl.ini
+#define CANITFTYPE_CAN_PEAK	0
+#define CANITFTYPE_CAN_PEAK_USN	1
```
mgruhler: Shouldn't this be CANITFTYPE_CAN_PEAK_USB?

mgruhler:
with exception of above comment, Visual Review :ok:

cob_generic_can/common/include/cob_generic_can/CanItf.h
floweisshardt: @destogl: please update typo

destogl:
Fixed, sorry for this!

floweisshardt:
visual review: OK

PR #172, opened 2014-10-22
Adds the stageprofi board for cob_light

thiagodefreitas:
- Some spaces/identation formatted due to issues with the old version

@ipa-nhg , @ipa-fmw Please test with the robot pc

### DIFF
--- a/cob_light/msg/LightMode.msg
+++ b/cob_light/msg/LightMode.msg
@@ -15,4 +15,5 @@ int32 pulses 			# spezifies the amount of pulses which will be executed.
 			 			# eg: mode = flash, pulses = 2. Meens the light will flash two times
 int8 priority 			# priority [-20,20] default = 0. Modes with same or higher priorities will 
 						# be executed.
-int32 led_number
+uint32[] led_numbers
+std_msgs/ColorRGBA[] colors
--- a/cob_light/msg/LightMode.msg
+++ b/cob_light/msg/LightMode.msg
@@ -15,3 +15,4 @@ int32 pulses 			# spezifies the amount of pulses which will be executed.
 			 			# eg: mode = flash, pulses = 2. Meens the light will flash two times
 int8 priority 			# priority [-20,20] default = 0. Modes with same or higher priorities will 
 						# be executed.
+int32 led_number
--- a/cob_light/ros/src/cob_light.cpp
+++ b/cob_light/ros/src/cob_light.cpp
@@ -75,7 +75,7 @@
 #include <cob_light/SetLightModeActionResult.h>
 
 // serial connection includes
- #include <serialIO.h>
+#include <serialIO.h>
 
 // additional includes
 #include <colorUtils.h>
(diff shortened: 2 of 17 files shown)


benmaidel:
@thiagodefreitas please remove all your tab to whitespace changes.

thiagodefreitas:
@ipa-bnm My editor does it automatically, as it was configured for other project using the [ROS C++ Style guidelines](http://wiki.ros.org/CppStyleGuide) and [roslint recommendations](http://wiki.ros.org/roslint), where no tabs should be used, and indentation should be taken as 2 spaces.

Does that cause problems to you,somehow? If so, I can revert it. But otherwise, I would suggest that we keep the changes.

benmaidel:
@thiagodefreitas No, this does not cause any problems on my side. The formating is ok. But the pull request is confusing because you can not see the actual implementation changes. It would be great if you could seperate the codeing style changes from the actual implementation changes into different pull request.

mgruhler:
@ipa-bnm one of the great hidden features of github to avoid this issue I stumbled upon a while ago: add ?w=1 to the end of the URL to get rid of whitespace-only changes in the diffview

benmaidel:
@ipa-mig hey, this is really nice :+1: But it is not possible to write source code comments in this view. I still think that seperating implementation from coding style is the right way to do it.

cob_light/msg/LightMode.msg
```diff
@@ -14,4 +14,5 @@ float32 timeout 		# in s, requested mode will be executed for max timout s.
 int32 pulses 			# spezifies the amount of pulses which will be executed.
 			 			# eg: mode = flash, pulses = 2. Meens the light will flash two times
 int8 priority 			# priority [-20,20] default = 0. Modes with same or higher priorities will 
-						# be executed.
\ No newline at end of file
+						# be executed.
+int32 led_number
```
benmaidel: i think that an array of color values will be more effective than setting each led separately. perhaps something like: std_msgs/ColorRGBA[] colors

cob_light/ros/src/cob_light.cpp
```diff
@@ -108,308 +109,330 @@ void shutdownCallback(XmlRpc::XmlRpcValue& params, XmlRpc::XmlRpcValue& result)
...
+      res.active_priority = p_modeExecutor->getExecutingPriority();
+      ROS_ERROR("Unsupported Color format. rgba values range is between 0.0 - 1.0");
+    }
+    else if(req.mode.mode == cob_light::LightMode::NONE)
+    {
+      p_modeExecutor->stop();
+      _color.a = 0;
+      if(_deviceDriver == "stageprofi")
+      {
+        std::cout << "Requested r" << req.mode.color.r;
```
benmaidel: unnecessary debugging output

cob_light/ros/src/cob_light.cpp
```diff
@@ -108,308 +109,330 @@ void shutdownCallback(XmlRpc::XmlRpcValue& params, XmlRpc::XmlRpcValue& result)
...
+    //ROS_DEBUG("Service Callback [Mode: %i with prio: %i freq: %f timeout: %f pulses: %i ]",
+    //	req.mode.mode, req.mode.priority, req.mode.frequency, req.mode.timeout, req.mode.pulses);
+
+    if(req.mode.color.r > 1.0 || req.mode.color.g > 1.0 || req.mode.color.b > 1.0 || req.mode.color.a > 1.0)
+    {
+      res.active_mode = p_modeExecutor->getExecutingMode();
+      res.active_priority = p_modeExecutor->getExecutingPriority();
+      ROS_ERROR("Unsupported Color format. rgba values range is between 0.0 - 1.0");
+    }
+    else if(req.mode.mode == cob_light::LightMode::NONE)
```
benmaidel, line 289: LightMode NONE => all Lights off

cob_light/ros/src/cob_light.cpp
```diff
@@ -108,308 +109,330 @@ void shutdownCallback(XmlRpc::XmlRpcValue& params, XmlRpc::XmlRpcValue& result)
...
+      p_modeExecutor->stop();
+      delete p_modeExecutor;
+    }
+    if(p_colorO != NULL)
+    {
+      delete p_colorO;
+    }
+  }
+
+  void topicCallback(std_msgs::ColorRGBA color)
```
benmaidel, line 258: i think stageprofi should also be handelt in this case?

cob_light/msg/LightMode.msg
thiagodefreitas: The possibility to control each LED separately is required for our use cases.

cob_light/ros/src/cob_light.cpp
thiagodefreitas, line 258: I may have misunderstood, but I think the p_color0 and the p_modeExecutor can be destroyed for the stageprofi there as well, without any additional if.

cob_light/ros/src/cob_light.cpp
thiagodefreitas, line 289: Fixed

cob_light/ros/src/cob_light.cpp
benmaidel, line 258: Yes, but i meant the topicCallback function. I think in this function the stageprofi needs to be handelt?

cob_light/msg/LightMode.msg
benmaidel: Then we should probably come up with a better solution. Calling x times the same action/service only to set all lights is horrible. Imagine if we have 512 leds or even more...
What about a index array and a value array?
uint32 [] led_numbers
std_msgs/ColorRGBA[] colors

cob_light/msg/LightMode.msg
```diff
@@ -14,4 +14,6 @@ float32 timeout 		# in s, requested mode will be executed for max timout s.
 int32 pulses 			# spezifies the amount of pulses which will be executed.
 			 			# eg: mode = flash, pulses = 2. Meens the light will flash two times
 int8 priority 			# priority [-20,20] default = 0. Modes with same or higher priorities will 
-						# be executed.
\ No newline at end of file
+						# be executed.
+uint32[] led_numbers
```
floweisshardt, line 18: @thiagodefreitas: please add some comment here to indicate for what led_numbers and colors can be used. (e.g. which mode, default value, ...). Should be similar to the comments for the other message entries above.

NadiaHG:
Visual Review: OK

floweisshardt:
visual review: OK

PR #176, opened 2014-11-17
Indigo new structure

floweisshardt:


NadiaHG:
Visual Review: OK

floweisshardt:
visual review: OK
tested in simulation
tested on cob4-2

mgruhler:
This has nothing to do with ros_control, right? So it does work with r@w and cob3 as well...

floweisshardt:
this was especially made for the "old" driver to be compatible with the new ros control plugins which are used in simulation too.

floweisshardt:
sammary of our discussion during lunch:
The goal is 
- to have a unique ros control plugin "omni_drive" which will substitute cob_undercarriage_ctrl for old and new hardware as well as for simulation
- to have an simpleIQ implementation in parallel to the canopen_402 implementation which will use the same 301 and ROS node layer as ros control plugin

floweisshardt:
visual review: OK
tested in simulation
tested on cob4-2

PR #178, opened 2014-11-18
new package cob_mimic - First Version

NadiaHG:


cob_mimic/msg/Mimic.msg
```diff
@@ -0,0 +1,13 @@
+string mimic
```
floweisshardt, line 1: I'd like to use a real string and not encode the string into ints. This means we can skip this message and only add a service "SetFace" which sends a string

cob_mimic/srv/SetFace.srv
```diff
@@ -0,0 +1 @@
+cob_mimic/Mimic face
```
floweisshardt, line 1: this is not a full service definition (missing reply message). please use 

req: std_msgs/String
res: EMPTY

Shouldn't we call this service SetMimic? (...because the package name is cob_minic and not cob_face)

cob_mimic/src/mimic_node.py
```diff
@@ -0,0 +1,266 @@
...
+        new.ptime = self.ptime
+        new.reversed = self.reversed
+        return new
+        
+def set_face(req):
+    pygame.init()
+    req_msg = Mimic
+    req_msg = req
+    print "Mimic: %s" % req_msg.face.mimic
+    file_localition = roslib.packages.get_pkg_dir('cob_mimic') + '/common/' + req_msg.face.mimic + '.gif'
```
floweisshardt, line 242: if we're using a real string in the SetFace service, we should check if the file exists. If not, return false.

cob_mimic/src/mimic_node.py
```diff
@@ -0,0 +1,266 @@
...
+    image = GIFImage(file_localition)
+    while 1:
```
floweisshardt, line 244: is this a good way to have a "loop true". We should at least add "and not rospy.is_shutdown()"

cob_mimic/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(cob_mimic)
+
+find_package(catkin REQUIRED COMPONENTS message_generation rospy actionlib actionlib_msgs std_msgs cob_srvs)
+
+#######################################
+## Declare ROS messages and services ##
+#######################################
+
+add_message_files(DIRECTORY msg FILES
```
floweisshardt, line 10: not needed anymore

cob_mimic/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+generate_messages(
+  DEPENDENCIES actionlib_msgs std_msgs
+)
+
+###################################
+## catkin specific configuration ##
+###################################
+
+catkin_package(
+    CATKIN_DEPENDS rospy actionlib actionlib_msgs std_msgs cob_srvs diagnostic_msgs message_runtime # TODO
```
floweisshardt, line 27: why do we need cob_srvs?, diagnostic_msgs, actionlib and actionlib_msgs?

cob_mimic/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+    INCLUDE_DIRS # TODO include
+    LIBRARIES # TODO
+)
+
+#############
+## Install ##
```
floweisshardt, line 33: we're missing to install the gifs

cob_mimic/package.xml
```diff
@@ -0,0 +1,37 @@
...
+  <url type="website">http://ros.org/wiki/cob_mimic</url>
+  <!-- <url type="bugtracker"></url> -->
+
+  <maintainer email="nhg@ipa.fhg.de">Nadia Hammoudeh Garcia</maintainer>
+  <author email="nhg@ipa.fhg.de">Nadia Hammoudeh Garcia</author>
+
+  <buildtool_depend>catkin</buildtool_depend>
+
+  <build_depend>rospy</build_depend>
+  <build_depend>actionlib</build_depend>
```
floweisshardt, line 18: why do we need cob_srvs?, diagnostic_msgs, actionlib and actionlib_msgs?

PR #189, opened 2014-12-04
cob_light updates

floweisshardt:


floweisshardt:
visual review: OK
tested on cob4-2: OK

Issue #171, opened 2014-10-08
cob_base_drive_chain or cob_undercarriage_control: if velocity twist message has NaN the wheels are uncontrollably rotate


destogl:
If velocity twist message has some of the values NaN this value is not filtered out but rather the wheels are starting uncontrollably rotate (around steering axis) and platform shuts down (probably because of to high current). This case should be avoided by checking the input values by corresponding components.

floweisshardt:
@destogl  thanks for pointing that out. Do you know which component was sending the NaN twist? That would be important to fix too.

destogl:
It was my component, it was a bug during develpement.

destogl:
Since is fixed I'll close it.

PR #197, opened 2015-01-23
bugfix: issue #171

ghost:
regarding bugfix of issue #171
to fix that bug: now checking for NaN-values in Twist message in undercarriage_cntrl node; if there are NaN-values, the robot is getting stopped and a ROS_FATAL message is getting send out;


mgruhler:
Visual Review :ok:
Tested on raw3-3 :ok:

floweisshardt:
Visual review: OK

PR #200, opened 2015-01-27
Bugfix issue #227

ghost:
regarding bugfix of issue #227
problem is: robot returns success in initialization of the platform, although emergency button is active and thus physically prevents performing homing procedure (motors signal communication error);
to fix that bug: rewrote relayboard_sim to improve simulation of the emergency stop state (toggeling of emergency button stop and scanner stop; also ROS_ERROR outputs); base_drive_chain node now subscribes to the emergency_stop_state topic; now, triggering the homing and recovering procedure is only possible, if the emergency_state in the cob_msgs/EmergencyStopState.msg is 0 (system is operating normal); if emergency_state is != 0, service calls for triggering homing and recovering return false;

regarding bugfix of issue #171 
to fix that bug: now checking for NaN-values in Twist message in undercarriage_cntrl node; if there are NaN-values, the robot is getting stopped and a ROS_FATAL message is getting send out;

### DIFF
--- a/cob_relayboard/ros/src/relayboard_sim.py
+++ b/cob_relayboard/ros/src/relayboard_sim.py
@@ -37,10 +37,9 @@ class RelayboardSim:
 		self.msg_voltage.data = 48.0 # in simulation battery is always full
 
 		# advertise services
-		srv = rospy.Service('~toggle_emergency_button_stop', Empty, relayboard_sim.toggle_emergency_button_stop)
-		srv = rospy.Service('~toggle_scanner_stop', Empty, relayboard_sim.toggle_scanner_stop)
-		srv = rospy.Service('~toggle_emergency_state', Empty, relayboard_sim.toggle_emergency_state)
-
+		srv = rospy.Service('~toggle_emergency_button_stop', Empty, self.toggle_emergency_button_stop)
+		srv = rospy.Service('~toggle_scanner_stop', Empty, self.toggle_scanner_stop)
+		
 	# service callback to simulate emergency button stop
 	def toggle_emergency_button_stop(self, req):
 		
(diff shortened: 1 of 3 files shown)


cob_relayboard/ros/src/relayboard_sim.py
```diff
@@ -5,38 +5,97 @@
...
+
+		self.msg_power_board.header.stamp = rospy.Time.now()
+		self.msg_power_board.run_stop = True
+		self.msg_power_board.wireless_stop = True # for cob the wireless stop field is misused as laser stop field
+
+		self.msg_voltage.data = 48.0 # in simulation battery is always full
+
+		# advertise services
+		srv = rospy.Service('~toggle_emergency_button_stop', Empty, self.toggle_emergency_button_stop)
+		srv = rospy.Service('~toggle_scanner_stop', Empty, self.toggle_scanner_stop)
```
mgruhler: give seperate names to those service servers

cob_relayboard/ros/src/relayboard_sim.py
```diff
@@ -5,38 +5,97 @@
...
+			else:
+				relayboard_sim.msg_em.emergency_state = 0
+				rospy.loginfo("Emergency stop released.")
+	
+		# publish topics
+		relayboard_sim.pub_em_stop.publish(relayboard_sim.msg_em)
+		relayboard_sim.pub_power_board.publish(relayboard_sim.msg_power_board)
+		relayboard_sim.pub_voltage.publish(relayboard_sim.msg_voltage)
+
+		rospy.sleep(0.5)
```
mgruhler: why do you set this to 0.5 seconds. before, it was 1.0?

mgruhler:
regarding ipa320/cob_robots#227

needs to be tested on robot prior to merge

floweisshardt:
I dont't think it's a good way to just subscribe to the emergency_stop topic. I'd like to avoid that and rather read out what the hardware tells us. The hardware layer should be able to tell us that homing wasn't possible due to undervoltage error or similiar.

checking for NAN values is OK. @ipa-mig-tk: please provide that in a separate PR so I can merge it without the changes with the emergency stop topic

ghost:
Ok. I'll handle that.

PR #202, opened 2015-02-02
bugfix regarding to ipa320/cob_driver#171: now checking for NaN-values in Twist message

ghost:
ipa320/cob_driver#171
to fix that bug: now checking for NaN-values in Twist message within undercarriage_cntrl node; if there are NaN-values, the robot is getting stopped and a ROS_FATAL message is send out;


mgruhler:
Visual Review :ok:
Tested on raw3-3 :ok:

floweisshardt:
visual review: OK

PR #203, opened 2015-02-02
bugfix regarding to ipa320/cob_robots#227

ghost:
ipa320/cob_robots#227
problem: robot returns success in initialization of the platform, although emergency button is active and thus physically prevents performing homing procedure (motors signal communication error);
to fix that bug: base_drive_chain node now subscribes to the emergency_stop_state topic; triggering the homing and recovering procedure is only possible, if the emergency_state in cob_msgs/EmergencyStopState.msg is 0 (system is operating normal); if emergency_state is != 0, service calls for triggering homing and recovering return false;


ghost:
ipa320/cob_robots#227
@ipa-fmw : regarding to your claim I'm now working on reading out the hardware layer to avoid subscribing to the emergency_stop_state topic; I leave this pull request open as long as I come up with a better matching solution;

ghost:
Closing pull request due to design demand "avoid subscribing to emergency state topic". I'll open a new pull request containing another solution.

PR #204, opened 2015-02-10
Bugfix ipa320/cob_robots#227

ghost:
bugfix ipa320/cob_robots#227
problem description: robot returns success in initialization of the platform although emergency button is active and thus physically prevents performing homing procedure (motors signal communication error);
to fix that bug (in ipa320/cob_command_tools/cob_teleop): now checking success of init/recover procedures instead of just checking success of service calls;
to fix that bug (in ipa320/cob_driver/base_drive_chain): changed wrong ROS output regarding base recover service callback

### DIFF
--- a/cob_base_drive_chain/ros/src/cob_base_drive_chain.cpp
+++ b/cob_base_drive_chain/ros/src/cob_base_drive_chain.cpp
@@ -430,7 +430,7 @@ class NodeClass
 		bool srvCallback_Init(cob_srvs::Trigger::Request &req,
 							  cob_srvs::Trigger::Response &res )
 		{
-			ROS_DEBUG("Service Callback init");
+			ROS_DEBUG("Service callback /base_controller/init");
 			if(m_bisInitialized == false)
 			{
 				m_bisInitialized = initDrives();
@@ -500,7 +500,7 @@ class NodeClass
 		{
 			if(m_bisInitialized)
 			{
-				ROS_DEBUG("Service callback reset");
+				ROS_DEBUG("Service callback /base_controller/recover");
 #ifdef __SIM__
 				res.success.data = true;
 #else
@@ -515,11 +515,10 @@ class NodeClass
 			}
 			else
 			{
-				ROS_WARN("...base already recovered...");
-				res.success.data = true;
-				res.error_message.data = "base already recovered";
+				ROS_WARN("Base not yet initialized.");
+				res.success.data = false;
+				res.error_message.data = "Base not yet initialized.";
 			}
-
 			return true;
 		}
 		


cob_base_drive_chain/ros/src/cob_base_drive_chain.cpp
```diff
@@ -515,11 +515,10 @@ class NodeClass
 			}
 			else
 			{
-				ROS_WARN("...base already recovered...");
-				res.success.data = true;
-				res.error_message.data = "base already recovered";
+				ROS_WARN("Base not yet initialized.");
+				res.success.data = false;
```
floweisshardt, line 519: why did you change the success behaviour? 
The `res.success.data` should contain the state of the base which means `true` if the base can be moved afterwards (recover successfully) and `false` if the base cannot be moved (recover failed).

floweisshardt, line 519: sorry, my mistake. looks good.

floweisshardt:
visual review: OK

cob_base_drive_chain/ros/src/cob_base_drive_chain.cpp
```diff
@@ -430,7 +430,7 @@ class NodeClass
 		bool srvCallback_Init(cob_srvs::Trigger::Request &req,
 							  cob_srvs::Trigger::Response &res )
 		{
-			ROS_DEBUG("Service Callback init");
+			ROS_DEBUG("Service callback /base_controller/init");
```
fmessmer, line 433: @ipa-mig-tk @ipa-mig 
even if this is only ROS_DEBUG, can we get rid of the hardcoded namespace `/base_controller` in the text - as this is going to changed anyway.
Without the hardcoded namespace we are old-new-namespace-independent...

cob_base_drive_chain/ros/src/cob_base_drive_chain.cpp
```diff
@@ -500,7 +500,7 @@ class NodeClass
 		{
 			if(m_bisInitialized)
 			{
-				ROS_DEBUG("Service callback reset");
+				ROS_DEBUG("Service callback /base_controller/recover");
```
fmessmer, line 503: @ipa-mig-tk @ipa-mig 
even if this is only ROS_DEBUG, can we get rid of the hardcoded namespace `/base_controller` in the text - as this is going to changed anyway.
Without the hardcoded namespace we are old-new-namespace-independent...

cob_base_drive_chain/ros/src/cob_base_drive_chain.cpp
mgruhler, line 433: @ipa-mig-tk Please take care of this. ipa-fxm has a point there. Didn't think of it when talking to you...
Just call it "Service init in cob_base_drive_chain" or something (same for below)

ghost:
Done. Sending a PR.

PR #215, opened 2015-04-14
diagnostics should be in global namespace

jobstvogt:


floweisshardt:
this was removed by purpose as it is best practice to have everything in local namespaces and the use remaps in the launch files to map it to global topics.

fmessmer:
Hmm, that's think about this a bit....
In case, there always only is a subscriber to (global) `/diagnostics`, this PR actually has a point....and I was too quick removing **all** global namespace slashes...

...but a remap in the [launch file](https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/drivers/head_axis.launch) should work, too

floweisshardt:
I think it's still cleaner to remove hardcoded global namespaces in the code and remap them during the deployment phase (e.g. in a launch file)

fmessmer:
The remap is done in https://github.com/ipa320/cob_robots/pull/294

PR #218, opened 2015-04-23
[sound, light] fix diagnostics

floweisshardt:


fmessmer:
The whole `main()` looks somewhat strange!
Especially [this part](https://github.com/ipa-cob4-2/cob_driver/blob/indigo_dev/cob_sound/ros/src/sound.cpp#L135-L140)
The main only instantiates an object of the `Say` class and then spins....
for publishing diagnostics we could create a ros::Timer and put the code for publishing into the timer_cb function....

floweisshardt:
+1
that's a todo for next wednesday

floweisshardt:
tested on cob4-2: OK

fmessmer:
fine with me

PR #227, opened 2015-06-08
Feature/scanner standby

mgruhler:
This PR introduces a check for whether the scanner is in standby into the driver.
This is signaled through a std_msgs::Bool topic.
If the scanner is in standby, it additionally publishes a throttled diagnostic msg.

The check is done by checking all scanner values for the bitflag 0x4004. It is the only way to check this without configuring the scanner differently and evaluating the I/O-protocol.

@ipa-josh


floweisshardt:
tested on cob4-2: OK

PR #230, opened 2015-07-27
New play service

NadiaHG:


cob_sound/action/Play.action
```diff
@@ -0,0 +1,6 @@
+# goal
+std_msgs/String filename
```
fmessmer, line 2: use `string filename`!

cob_sound/action/Play.action
```diff
@@ -0,0 +1,6 @@
...
+---
+#result
```
fmessmer, line 4: can we have a

```
bool success
string message
```

in both feedback and result?

cob_sound/ros/src/sound.cpp
```diff
@@ -6,6 +6,8 @@
 #include <std_srvs/Trigger.h>
 #include <cob_sound/SayAction.h>
 #include <cob_sound/SayText.h>
+#include <cob_sound/PlayAction.h>
+#include <cob_sound/PlayFile.h>
 
 class SayAction
```
fmessmer, line 12: this is no longer just a `SayAction`...can we rename it to something like `SoundAction` or `SoundServer`?

cob_sound/ros/src/sound.cpp
```diff
@@ -17,6 +19,7 @@ class SayAction
   ros::ServiceServer srvServer_mute_;
   ros::ServiceServer srvServer_unmute_;
   ros::Subscriber sub_;
```
fmessmer, line 21: can we suffix this to `sub_say_`?

cob_sound/ros/src/sound.cpp
```diff
@@ -32,9 +35,11 @@ class SayAction
   {
     as_.start();
     srvServer_ = nh_.advertiseService("say", &SayAction::service_cb, this);
+    srvServer_ = nh_.advertiseService("play", &SayAction::service_cb_play, this);
```
fmessmer, line 38: can the same ServiceServer advertise two services?
or do we need two ServiceServers - one for "say" and one for "play"?

cob_sound/ros/src/sound.cpp
```diff
@@ -17,6 +19,7 @@ class SayAction
   ros::ServiceServer srvServer_mute_;
   ros::ServiceServer srvServer_unmute_;
   ros::Subscriber sub_;
+  ros::Subscriber sub_play_;
   std::string action_name_;
```
fmessmer, line 23: The `SimpleActionServer<cob_sound::PlayAction>` is missing!

NadiaHG:
Actually we can rewrite the node, I am also not sure that we need topic, service and action server for each functionality, and we can define just a generic Sound.action and sound.srv. What do you think @ipa-fmw ? Maybe cob_sound should have some consistence with cob_mimic?

cob_sound/action/Play.action
floweisshardt, line 4: what would you use it for?

floweisshardt:
+1 for the suggestions from @ipa-fxm

I'd like to keep text-to-speech (say) separate from playing audio files (play), so two separate messages are ok for me (no generic sound.action or sound.srv). 

We should also keep the topic, service and action interfaces.

floweisshardt:
I just discussed with @ipa-nhg: 
- we'll use cob_srvs/SetString for both say and play
- we'll keep all three interfaces (topic, service, action)
- we'll introduce a mandatory audio_file_path parameter and use it to compose full path (audio_file_path + / + filename + .wav)
- we'll have to modify cob_script_server to use the action interface for play

fmessmer:
Fine...awaiting a PR update to review on Monday :wink:

cob_sound/action/Play.action
fmessmer, line 4: If we can evaluate in the ActionServer, whether the command has successfully been executed / or failed due to a certain failure...I guess it would be nice to send this in the result...just because this is the purpose of the result part of the message...

NadiaHG:
New PR https://github.com/ipa320/cob_driver/pull/231
I will close this one

PR #231, opened 2015-07-29
play sound

NadiaHG:


cob_sound/action/Play.action
```diff
@@ -0,0 +1,6 @@
+# goal
+string filename
+---
+#result
```
fmessmer, line 4: so no result?

cob_sound/ros/src/sound.cpp
```diff
@@ -5,18 +5,22 @@
...
+  actionlib::SimpleActionServer<cob_sound::SayAction> as_say_;
+  actionlib::SimpleActionServer<cob_sound::PlayAction> as_play_;
+  ros::ServiceServer srvServer_say_;
+  ros::ServiceServer srvServer_play_;
   ros::ServiceServer srvServer_mute_;
   ros::ServiceServer srvServer_unmute_;
-  ros::Subscriber sub_;
+  ros::Subscriber sub_say_;
+  ros::Subscriber sub_play_;
   std::string action_name_;
```
fmessmer: this is obsolete ad you coded the ActionName into the advertise directly...
...or we would need two strings - one for each action

cob_sound/ros/src/sound.cpp
```diff
@@ -26,48 +30,67 @@ class SayAction
...
-    srvServer_unmute_ = nh_.advertiseService("unmute", &SayAction::service_cb_unmute, this);
-    sub_ = nh_.subscribe("say", 1000, &SayAction::topic_cb, this);
+    as_say_.start();
+    as_play_.start();
+    srvServer_say_ = nh_.advertiseService("say", &SoundAction::service_cb_say, this);
+    srvServer_play_ = nh_.advertiseService("play", &SoundAction::service_cb_play, this);
+    srvServer_mute_ = nh_.advertiseService("mute", &SoundAction::service_cb_mute, this);
+    srvServer_unmute_ = nh_.advertiseService("unmute", &SoundAction::service_cb_unmute, this);
+    sub_say_ = nh_.subscribe("say", 1000, &SoundAction::topic_cb_say, this);
+    sub_play_ = nh_.subscribe("play", 1000, &SoundAction::topic_cb_play, this);
```
fmessmer: queue  size better be 1

cob_sound/ros/src/sound.cpp
```diff
@@ -130,7 +163,45 @@ class SayAction
...
+    }
+    if(!nh_.hasParam("audio_file_path"))
+    {
+      ROS_ERROR("Parameter 'audio_file_path' not set");
+      return false;
+    }
+    std::string command;
+    std::string audio_file_path;
+    nh_.getParam("audio_file_path",audio_file_path);
+    std::string audio_file= audio_file_path +"/"+filename+".wav";
```
fmessmer: Few comments here:
- Why do we check for parameter existence every `play()`. Can we not have this in the constructor/init as a member variable?
- I'd better like if we could specify `audio_file_path` in a `${find package_name}` fasion
- Can we only play `*.wav`? I'd prefer to not have the file-suffix added hardcoded like this. I'd rather expect the user to fully name the filename **with** suffix. If we can only play `*.wav`'s we could reject and return `false`.

cob_sound/ros/src/sound.cpp
```diff
@@ -130,7 +163,45 @@ class SayAction
...
+      status.message = "command play failed";
+      diagnostics_.status.push_back(status);
+      
+      diagnostics_.header.stamp = ros::Time::now();
+      diagnostics_pub_.publish(diagnostics_);
+      
+      diagnostics_.status.resize(0);
+      return false;
+    }
+    return false;
```
fmessmer: shoud this not return `true` here?

fmessmer:
From the discussion with @ipa-nhg @ipa-fmw:
`cob_sound` receives fully qualified path name, i.e. e.g. "/home/robot/sound/happy.wav".
In `cob_script_server`, a convenient functiion will be provided:

```
def play_sound(self, file_name, file_path="", file_suffix=".wav")
```

If file_path is "", it will be replaced by a `default_audio_path` parameter, which can be set in `cob_robot_default_config`

cob_sound/ros/src/sound.cpp
```diff
@@ -26,48 +29,66 @@ class SayAction
   ros::Timer diagnostics_timer_;
   ros::Publisher pubMarker_;
 
-  SayAction(std::string name) :
-    as_(nh_, name, boost::bind(&SayAction::as_cb, this, _1), false),
-    action_name_(name)
+  SoundAction(std::string name) :
```
floweisshardt: name is not used any more, can be removed

cob_sound/ros/src/sound.cpp
```diff
@@ -181,7 +242,7 @@ int main(int argc, char** argv)
 {
   ros::init(argc, argv, "cob_sound");
 
-  SayAction say("say");
+  SoundAction sound("sound");
```
floweisshardt: delete "sound"

cob_sound/action/Play.action
```diff
@@ -0,0 +1,6 @@
+# goal
+string filename
```
floweisshardt: add a comment:
- this is the fully qualified filename including file ending, example: "/home/user/sounds/hello.wav"

floweisshardt:
LGTM

cob_sound/action/Say.action
```diff
@@ -1,5 +1,5 @@
 # goal
-std_msgs/String text
+string text
```
fmessmer, line 2: as now none of the messages requires `std_msgs` in its definition anymore, please remove `std_msgs` from https://github.com/ipa-nhg/cob_driver/blob/95181a8b5b666d5e76c77464a3513465c5d79cd1/cob_sound/CMakeLists.txt#L13

floweisshardt, line 2: done in https://github.com/ipa320/cob_driver/pull/233

PR #247, opened 2015-11-16
fix frame_id for odometry

floweisshardt:


NadiaHG:
Visual Review : OK
Test on cob4-2: OK

mgruhler:
:ok: however, we should occasionally make this a parameter...

fmessmer:
Also, according to http://wiki.ros.org/tf2/Migration#Removal_of_support_for_tf_prefix frames starting with `/` will not be allowed for tf2...
Thus, we might want to start migrating whenever we stumble over s.th. lke this

floweisshardt:
the `/` problem should be discussed together with the other issues from the multi-robot experients

PR #239, opened 2015-09-25
[WIP] Allow multi-robot simulation

ipa-srd-fg:
- cob_relayboard delete slashes topics
- This PR should only be merged with all the other multi-robot-related PRs...and after more testing!
  
  ipa320/cob_common#168
  ipa320/cob_control#55
  ipa320/cob_robots#331
  ipa320/cob_simulation#92
  ipa320/ipa_navigation_localization#99
  ipa320/ipa_navigation#80

PR #250, opened 2015-11-20
Restructure relayboard nodehandles

fmessmer:
This PR remove hard-coded global namespaces within the node itself...which we always should try to avoid...It was taken from #239 separating this feature from the scan remap which is not needed anymore due to https://github.com/ipa320/cob_robots/pull/376
For our use case within [relayboard.launch](https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/drivers/relayboard.launch) the topics will not change at all for our robots

### DIFF
--- a/cob_relayboard/ros/src/cob_relayboard_node.cpp
+++ b/cob_relayboard/ros/src/cob_relayboard_node.cpp
@@ -99,7 +99,7 @@ public:
 
     topicPub_isEmergencyStop = n.advertise<cob_msgs::EmergencyStopState>("emergency_stop_state", 1);
     topicPub_PowerBoardState = n.advertise<cob_msgs::PowerBoardState>("power_board/state", 1);
-    topicPub_Voltage = n.advertise<std_msgs::Float64>("power_board/voltage", 10);
+    topicPub_Voltage = n.advertise<std_msgs::Float64>("power_board/voltage", 1);
 
     // Make sure member variables have a defined state at the beginning
     EM_stop_status_ = ST_EM_FREE;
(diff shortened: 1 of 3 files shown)


cob_relayboard/ros/src/cob_relayboard_node.cpp
```diff
@@ -93,11 +94,12 @@ class NodeClass
...
-    n = ros::NodeHandle("~");
+    n = ros::NodeHandle();
+    n_priv = ros::NodeHandle("~");
 
-    topicPub_isEmergencyStop = n.advertise<cob_msgs::EmergencyStopState>("/emergency_stop_state", 1);
-    topicPub_PowerBoardState = n.advertise<cob_msgs::PowerBoardState>("/power_board/state", 1);
-    topicPub_Voltage = n.advertise<std_msgs::Float64>("/power_board/voltage", 10);
+    topicPub_isEmergencyStop = n.advertise<cob_msgs::EmergencyStopState>("emergency_stop_state", 1);
+    topicPub_PowerBoardState = n.advertise<cob_msgs::PowerBoardState>("power_board/state", 1);
+    topicPub_Voltage = n.advertise<std_msgs::Float64>("power_board/voltage", 10);
```
benmaidel: why queue size 10? `power_board/state` and `emergency_stop_state` are in the same update loop

fmessmer: we took that as it has been...but it clearly makes sense to harmonize his...done!
(+ tabs vs. spaces)

mgruhler:
CR: :ok:

NadiaHG:
CR: OK 
already tested on a robot?

floweisshardt:
CR: OK

PR #251, opened 2015-11-24
adjust filter topics

fmessmer:
This simplifies the confusion in https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/drivers/laser_scan_filter.launch and is conform to the new naming convention


mgruhler:
CR: :ok:

floweisshardt:
CR: OK

PR #257, opened 2016-01-07
Update scan_unifier to use tf2

destogl:


cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -295,8 +298,8 @@ sensor_msgs::LaserScan scan_unifier_node::unifieLaserScans()
         {
           // use the nearest reflection point of all scans for unified scan
           unified_scan.ranges.at(index) = sqrt(range_sq);
-          // get respective intensity from point cloud intensity-channel (index 0)
-          unified_scan.intensities.at(index) = vec_cloud.at(j).channels.at(0).values.at(i);
+          // get respective intensity from point cloud intensity-channel
+		  unified_scan.intensities.at(index) = vec_cloud.at(j).points[i].intensity;
```
fmessmer: indentation (tabs vs. spaces)

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -230,25 +233,25 @@ void scan_unifier_node::topicCallbackLaserScan(const sensor_msgs::LaserScan::Con
 sensor_msgs::LaserScan scan_unifier_node::unifieLaserScans()
 {
   sensor_msgs::LaserScan unified_scan = sensor_msgs::LaserScan();
-  std::vector<sensor_msgs::PointCloud> vec_cloud;
-  vec_cloud.assign(config_.number_input_scans, sensor_msgs::PointCloud());
+  std::vector<sensor_msgs::PointCloud2> temp_cloud;
+  temp_cloud.assign(config_.number_input_scans, sensor_msgs::PointCloud2());
+  std::vector<Cloud> vec_cloud;
```
fmessmer: is this `pcl::Cloud`? 
Can we use explicit namespace, please

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -230,25 +233,25 @@ void scan_unifier_node::topicCallbackLaserScan(const sensor_msgs::LaserScan::Con
...
+  vec_cloud.assign(config_.number_input_scans, Cloud());
```
fmessmer: explicit namespace

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -230,25 +233,25 @@ void scan_unifier_node::topicCallbackLaserScan(const sensor_msgs::LaserScan::Con
...
       ROS_DEBUG_STREAM("Converting scans to point clouds at index: " << i << ", at time: " << vec_laser_struct_.at(i).current_scan_msg.header.stamp << " now: " << ros::Time::now());
       try
       {
-        listener_.waitForTransform("/base_link", vec_laser_struct_.at(i).current_scan_msg.header.frame_id,
-            vec_laser_struct_.at(i).current_scan_msg.header.stamp, ros::Duration(3.0));
-
         ROS_DEBUG("now project to point_cloud");
-        projector_.transformLaserScanToPointCloud("/base_link",vec_laser_struct_.at(i).current_scan_msg, vec_cloud.at(i), listener_);
+        projector_.transformLaserScanToPointCloud("base_link", vec_laser_struct_.at(i).current_scan_msg, temp_cloud.at(i), *p_tfBuffer);
+        pcl::fromROSMsg<Point>(temp_cloud.at(i), vec_cloud.at(i));
```
fmessmer: explicit namespace

cob_scan_unifier/src/scan_unifier_node.cpp
mgruhler: there are typedefs in the header: line 89f

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -230,25 +233,25 @@ void scan_unifier_node::topicCallbackLaserScan(const sensor_msgs::LaserScan::Con
...
     for(int i=0; i < config_.number_input_scans; i++)
     {
-      vec_cloud.at(i).header.stamp = vec_laser_struct_.at(i).current_scan_msg.header.stamp;
+      temp_cloud.at(i).header.stamp = vec_laser_struct_.at(i).current_scan_msg.header.stamp;
       ROS_DEBUG_STREAM("Converting scans to point clouds at index: " << i << ", at time: " << vec_laser_struct_.at(i).current_scan_msg.header.stamp << " now: " << ros::Time::now());
       try
       {
-        listener_.waitForTransform("/base_link", vec_laser_struct_.at(i).current_scan_msg.header.frame_id,
-            vec_laser_struct_.at(i).current_scan_msg.header.stamp, ros::Duration(3.0));
-
```
mgruhler, line 247: why do you remove this? (Could discuss, whether this makes sense, but is there any particular reason?

mgruhler:
needs to be tested on HW.

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -71,6 +71,9 @@ scan_unifier_node::scan_unifier_node()
 
   getParams();
 
+  p_tfBuffer = new tf2_ros::Buffer();
+  p_tfListener = new tf2_ros::TransformListener(*p_tfBuffer, true);
```
benmaidel: please free allocated memory or use smart pointers

cob_scan_unifier/src/scan_unifier_node.cpp
destogl, line 247: Because waitForTransform doesn't exist in tf2 (see: [ROS wiki](http://wiki.ros.org/tf2/Tutorials/tf2%20and%20time%20%28C%2B%2B%29))

mgruhler, line 247: Ah :-) Didn't think of this...

Hmm, I guess the intention of this was to make sure that the transform is there so the `transformLaserScanToPointCloud` does not throw an exception. Do you know how this is handled therein?

destogl:
I tested with hardware and works. But of course would be grate if someone else also testes.

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -230,26 +233,26 @@ void scan_unifier_node::topicCallbackLaserScan(const sensor_msgs::LaserScan::Con
...
-
         ROS_DEBUG("now project to point_cloud");
-        projector_.transformLaserScanToPointCloud("/base_link",vec_laser_struct_.at(i).current_scan_msg, vec_cloud.at(i), listener_);
+        projector_.transformLaserScanToPointCloud("base_link", vec_laser_struct_.at(i).current_scan_msg, temp_cloud.at(i), *p_tfBuffer.get());
+        pcl::fromROSMsg<pcl::PointXYZI>(temp_cloud.at(i), vec_cloud.at(i));
       }
-      catch(tf::TransformException ex){
-        ROS_ERROR("%s",ex.what());
+      catch(tf2::TransformException ex){
+        //ROS_ERROR("%s",ex.what());
```
fmessmer, line 255: why not keep the `ROS_ERROR`?

destogl, line 255: Because I get too much output. The thing is that underlying function `transformLaserScanToPointCloud` uses `ros::time::now()` and since if often needed "extrapolation in the future" there are errors. With tf is the same problem. So not to have full screen of errors I commented it out.

I know this is not optimal solution but I don't have better idea...

fmessmer, line 255: this is why we had the `waitForTransform` in `tf`..if this is not working in `tf2` I suggest to stick with `tf`!

destogl, line 255: This would be O.K. but tf is already deprecated for a long time, and  should soon be kicked out. Anyway it is O.K. to stick with tf. In the mean time I will change underlying library to support "waiting" in tf2 since there is an [easy solution](http://wiki.ros.org/tf2/Tutorials/tf2%20and%20time%20%28C%2B%2B%29).

fmessmer, line 255: Thinking about it...there **needs** to be a mechanism to deal with this extrapolation issue within tf2, too, right? Or we need to take care of proper TimeStamps being used....

fmessmer, line 255: Seems like tf2's lookupTransform, includes the waitForTransform functionality: see [this tutorial](http://wiki.ros.org/tf2/Tutorials/tf2%20and%20time%20%28C%2B%2B%29)
So, it seems that as long as [laser_geometry](http://docs.ros.org/jade/api/laser_geometry/html/classlaser__geometry_1_1LaserProjection.html) doesn't use the tf2 variants of lookupTransform, i.e. provide a `transformLaserScanToPointCloud` with a timeout for example, we cannot do anything about it...

destogl, line 255: Exactly, that is why I meant to do library changes

ipa-fez:
@ipa-fxm stale/close?

destogl:
It is still relevant, but I didn't have time to solve it. Hopefully should be in the next months.

Nevertheless it could be closed and I will provide new one with up-to date changes.

ipa-fez:
Closing in favour of new up-to-date PR.

PR #259, opened 2016-01-21
use aggregated power state message

floweisshardt:


benmaidel:
CR: OK

fmessmer:
Is this related to https://github.com/ipa320/cob_robots/issues/403?
Will it affect any other robot?
If Publishers have been removed, should we also find and remove Subscribers (in other nodes) that subscribed to the removed topics, too?

PRs have the same name, but they are not cross-referenced:
https://github.com/ipa320/cob_common/pull/182
https://github.com/ipa320/cob_command_tools/pull/129

fmessmer:
as discussed...and travis is happy...merging

PR #261, opened 2016-02-16
Fix mimic shutdown

fmessmer:
This PR fixes the shutdown behavior of cob_mimic....before it always needed to be `SIGTERM`'ed
There were some strange `while`-loops and `sleeps` in the code
It seems that `repeat=0` was never a proper option, thus, I removed this feature

It also replaces all `print` with proper `rospy.logXXX`
It also fixes the service name to `/mimic/set_mimic` and the action name in the test client
It also adds a proper response to the service

### DIFF
--- a/cob_mimic/src/mimic_node.py
+++ b/cob_mimic/src/mimic_node.py
@@ -70,79 +70,60 @@ from cob_mimic.msg import *
 
 class Mimic:
   def service_cb(self, req):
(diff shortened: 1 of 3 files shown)


fmessmer:
Tested on cob4-1

fmessmer:
re-adding the copying of video files after talking to @ipa-fmw

floweisshardt:
FR: OK

PR #262, opened 2016-02-16
cob_light new features

benmaidel:
new features:
- pause / resume modes if mode with higher prio is requested / ended
- service / action returns a tracking id, with which the mode can be stopped
- new service `stopLightMode`
- changed command topic msg from `std_msgs/colorRGBA` to `cob_msgs/colorRGBAArray`
- separated mode types definition and mode message definition into separate messages
- new glow mode

### DIFF
--- a/cob_light/common/include/glowColorMode.h
+++ b/cob_light/common/include/glowColorMode.h
@@ -1,0 +1,120 @@
+/****************************************************************
+ *
+ * Copyright (c) 2016
+ *
+ * Fraunhofer Institute for Manufacturing Engineering
+ * and Automation (IPA)
+ *
+ * +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+ *
+ * Project name: care-o-bot
+ * ROS stack name: cob_driver
+ * ROS package name: cob_light
+ * Description: Switch robots led color by sending data to
+ * the led-µC over serial connection.
+ *
+ * +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+ *
+ * Author: Benjamin Maidel, email:benjamin.maidel@ipa.fraunhofer.de
+ * Supervised by: Benjamin Maidel, email:benjamin.maidel@ipa.fraunhofer.de
+ *
+ * Date of creation: Feb 2016
+ * ToDo:
+ *
+ * +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of the Fraunhofer Institute for Manufacturing
+ *       Engineering and Automation (IPA) nor the names of its
+ *       contributors may be used to endorse or promote products derived from
+ *       this software without specific prior written permission.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License LGPL as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License LGPL for more details.
+ *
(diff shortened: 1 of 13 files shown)


cob_light/msg/LightModes.msg
```diff
@@ -0,0 +1,13 @@
...
+uint8 BREATH =         3           # will change the LEDs smoothly with "frequency" from "color" to black
+uint8 BREATH_COLOR =   4           # will change the LEDs smoothly with "frequency" from "color" to black
+                                   # and flips color in time
+uint8 FADE_COLOR =     5           # will fade the colors in rainbow
+uint8 SEQ =            6           # executes one sequence after another as defined in sequences array
+uint8 CIRCLE_COLORS =  7           # circle through specific colors, if one color is set only one color will circle
+                                   # if more than one color is set, that this colors will circle through
+uint8 SWEEP =          8           # circle color from front to back on both sides
+uint8 DIST_APPROX =    9           # color led circle from green to red depended on the measured laserscan dists
+uint8 GLOW =          10           # glow the color
```
fmessmer, line 13: Why did you separate this from the `LightMode` message?
This is not a message that is meant to be sent on a topic....and thus confusing...

cob_light/ros/src/cob_light.cpp
```diff
@@ -191,12 +195,12 @@ class LightControl
     //Subscribe to LightController Command Topic
     _sub = _nh.subscribe("command", 1, &LightControl::topicCallback, this);
 
-    //Subscribe to LightController Command Topic
-    _sub_mode = _nh.subscribe("light", 1, &LightControl::topicCallbackMode, this);
-
```
fmessmer, line 196: why did you remove this subscriber?
what is/was the difference between `light` and `command`?

cob_light/common/include/glowColorMode.h
```diff
@@ -0,0 +1,120 @@
...
+ * If not, see <http://www.gnu.org/licenses/>.
+ *
+ ****************************************************************/
+
+#ifndef GLOWCOLORMODE_H
+#define GLOWCOLORMODE_H
+
+#include <mode.h>
+
+class GlowColorMode : public Mode
```
fmessmer, line 60: just out of couriosity:
is there a documentation about the modes available....and - more importantly -  **what each mode is doing?** :wink:

cob_light/common/include/modeExecutor.h
```diff
@@ -67,26 +68,31 @@ class ModeExecutor
...
-	void execute(cob_light::LightMode requestMode);
+	uint64_t execute(boost::shared_ptr<Mode> mode);
+	uint64_t execute(cob_light::LightMode requestMode);
 
 	int getExecutingPriority();
 	int getExecutingMode();
+	uint64_t getExecutingUId();
 
+	void pause();
+	void resume();
```
fmessmer, line 79: `pause` and `resume` is something the light driver handles internally only, right?
Depending on the modes that have been started and their respective priority...
It is not exposed to the user, correct?

cob_light/ros/src/cob_light.cpp
```diff
@@ -191,12 +195,12 @@ class LightControl
...
     //Advertise light mode Service
     _srvServer = _nh.advertiseService("set_light", &LightControl::serviceCallback, this);
 
+    //Advertise stop mode service
+    _srvStopMode = _nh.advertiseService("stop_mode", &LightControl::stopMode, this);
```
fmessmer, line 202: does a cancle of the `set_light` action do the same as a call to the `stop_light` service?

cob_light/msg/LightModes.msg
benmaidel, line 13: I've separated them because we will not get md5 checksum error if we extend the LightModes with a new mode.

cob_light/ros/src/cob_light.cpp
benmaidel, line 196: because i've removed the possibility to switch modes via topic. This functionally was deprecated for a long time

cob_light/common/include/glowColorMode.h
benmaidel, line 60: Nope, you can take a look inside the LightModes.msg. There you will find some explanations on the different modes

cob_light/common/include/modeExecutor.h
benmaidel, line 79: correct. The user is only able to start new modes with a respective priority and stop running modes if he tracked them

cob_light/ros/src/cob_light.cpp
benmaidel, line 202: nope, because the action will return succeeded as soon as the requested mode started

benmaidel:
successfully tested on raw3-3 and cob4-2

cob_light/ros/src/cob_light.cpp
fmessmer, line 202: so we  will not be able to call the stop functionality from a action client interface....just for the record...

fmessmer:
FR ok

PR #265, opened 2016-03-01
cob_sound: use libvlc for audio playback

benmaidel:
features:
- play sound files with libvlc
- canceling action will stop the sound
- action publish feedback (position, time)
- added service to stop sound play

### DIFF
--- a/cob_sound/package.xml
+++ b/cob_sound/package.xml
@@ -10,6 +10,7 @@
 
   <maintainer email="nhg@ipa.fhg.de">Nadia Hammoudeh Garcia</maintainer>
   <author>Florian Weisshardt</author>
+  <author>Benjamin Maidel</author>
 
   <buildtool_depend>catkin</buildtool_depend>
 
@@ -20,6 +21,7 @@
   <depend>actionlib</depend>
   <depend>cob_srvs</depend>
   <depend>diagnostic_msgs</depend>
+  <depend>libvlc-dev</depend>
   <depend>roscpp</depend>
   <depend>std_msgs</depend>
   <depend>std_srvs</depend>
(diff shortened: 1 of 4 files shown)


cob_sound/package.xml
```diff
@@ -20,6 +21,7 @@
   <depend>actionlib</depend>
   <depend>cob_srvs</depend>
   <depend>diagnostic_msgs</depend>
+  <depend>libvlc-dev</depend>
```
fmessmer, line 24: needs to be proposed as dependency in a PR to rosdistro...@ipa-bnm has prepared it

cob_sound/ros/src/sound.cpp
```diff
@@ -91,6 +116,28 @@ class SoundAction
...
+    }
+    else
+    {
+      if(libvlc_media_player_is_playing(vlc_player_) == 1)
+      {
+        libvlc_media_player_stop(vlc_player_);
+        ret = true;
+      }
+    }
+    return ret;
```
fmessmer: as discussed:
- maybe always return true in order to indicate that server has been called successfully
- use res.succes = false to indicate that request could not be processed, i.e. because of mute

apply for all service callbacks...

cob_sound/ros/src/sound.cpp
```diff
@@ -165,17 +212,24 @@ class SoundAction
 
   bool play(std::string filename)
```
fmessmer, line 219: use this for res.succes instead of return of service callback

cob_sound/package.xml
benmaidel, line 24: PR is open https://github.com/ros/rosdistro/pull/10591

fmessmer:
CR ok

merge once rosdistro PR: https://github.com/ros/rosdistro/pull/10591 is merged (and travis happy)

floweisshardt:
FR: OK

PR #267, opened 2016-03-03
fix sound dependencies

floweisshardt:
fixes missing dependency from #265

### DIFF
--- a/cob_sound/package.xml
+++ b/cob_sound/package.xml
@@ -26,6 +26,7 @@
   <depend>std_msgs</depend>
   <depend>std_srvs</depend>
   <depend>visualization_msgs</depend>
+  <depend>vlc</depend>
 
   <exec_depend>alsa-oss</exec_depend>
   <exec_depend>rospy</exec_depend>


cob_sound/package.xml
```diff
@@ -26,6 +26,7 @@
   <depend>std_msgs</depend>
   <depend>std_srvs</depend>
   <depend>visualization_msgs</depend>
+  <depend>vlc</depend>
```
fmessmer, line 29: why do we need `<depend>vlc</depend>` if we have the `<depend>libvlc-dev</depend>` defined?

floweisshardt, line 29: we're seeing the following error:

```
process[sound/sound_node-1]: started with pid [3058]
[0x93e97e0] main libvlc error: No plugins found! Check your VLC installation.
[sound/sound_node-1] process has died [pid 3058, exit code -11, cmd devel/lib/cob_sound/sound diagnostics:=/diagnostics __name:=sound_node __log:=~/.ros/log/1b06cea6-e9ca-11e5-bd3f-eca86bff61c1/sound-sound_node-1.log].
```

NadiaHG:
CR: OK

fmessmer:
fine, then merging....

PR #275, opened 2016-04-08
Issues cob sound

floweisshardt:
superceeds #274

### DIFF
--- a/cob_sound/ros/src/sound.cpp
+++ b/cob_sound/ros/src/sound.cpp
@@ -73,10 +73,18 @@ public:
   void as_goal_cb_play_()
   {
     std::string filename = as_play_.acceptNewGoal()->filename;
-    if(play(filename))
+    std::string message;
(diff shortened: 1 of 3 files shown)


cob_sound/ros/src/sound.cpp
```diff
@@ -41,6 +41,7 @@ class SoundAction
     as_say_(nh_, "say", boost::bind(&SoundAction::as_cb_say_, this, _1), false),
     as_play_(nh_, "play", false)
   {
+    nh_ = ros::NodeHandle("~");
```
fmessmer, line 44: This will modify the namespace of the sound services and topic, right?
Will this have an effect on simple script server and other packages?

floweisshardt, line 44: no, the namespace is set in cob_bringup/sound.launch (equivalent to cob_light)

PR #278, opened 2016-07-02
calculate remaining capacity form voltage

timhappy:
canged calculation for remaining capacity on the basis on the voltage level. Should be removed if bms-problems are fixed.

### DIFF
--- a/cob_bms_driver/src/power_state_aggregator.py
+++ b/cob_bms_driver/src/power_state_aggregator.py
@@ -19,7 +19,7 @@ class PowerStateAggregator():
         self.remaining_capacity = None
         self.full_charge_capacity = None
         self.temperature = None
-        self.threshold_voltage_warning = rospy.get_param("~threshold_voltage_warning", 43) # voltage of battery level
+        self.threshold_voltage_warn = rospy.get_param("~threshold_voltage_warn", 43) # voltage of battery level
         self.threshold_voltage_error = rospy.get_param("~threshold_voltage_error", 41.0)     # voltage of battery level
         self.threshold_voltage_critical = rospy.get_param("~threshold_voltage_critical", 40)# voltage of battery level
         self.threshold_voltage_mid = rospy.get_param("~threshold_voltage_mid", 48)  # voltage of battery level
@@ -81,9 +81,15 @@ class PowerStateAggregator():
                 self.relative_remaining_capacity = 50 # % of full capacity
             else: #self.voltage <= self.threshold_voltage_full and self.current < self.steady_current:
                 self.relative_remaining_capacity = 100 # % of full capacity
+
+            return self.relative_remaining_capacity
+
         else:
             rospy.logwarn("cannot calculate relative remaining capacity. full_charge_capacity=%d, remaining_capacity=%d" % (self.full_charge_capacity, self.remaining_capacity))
             return 0.0
+
+        return self.relative_remaining_capacity
+
 
     def calculate_time_remaining(self):
         if len(self.last_currents) > 0:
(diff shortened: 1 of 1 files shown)


cob_bms_driver/src/power_state_aggregator.py
```diff
@@ -63,11 +71,26 @@ def calculate_power_consumption(self):
...
     def calculate_relative_remaining_capacity(self):
         if self.full_charge_capacity != None and self.remaining_capacity != None:
-            return round(100.0*(self.remaining_capacity/self.full_charge_capacity), 3)
+            if self.voltage <= self.threshold_voltage_critical and self.current > -self.steady_current:
+                self.relative_remaining_capacity = 5 # % of full capacity
+            elif self.voltage <= self.threshold_voltage_error and self.current > -self.steady_current:
+                self.relative_remaining_capacity = 10 # % of full capacity
+            elif self.voltage <= self.threshold_voltage_warn and self.current > -self.steady_current:
+                self.relative_remaining_capacity = 15 # % of full capacity
+            elif self.voltage <= self.threshold_voltage_mid and self.current > -self.steady_current:
```
mgruhler, line 80: why do you compare against negative `steady_current`?

cob_bms_driver/src/power_state_aggregator.py
```diff
@@ -63,11 +71,26 @@ def calculate_power_consumption(self):
...
+            else: #self.voltage <= self.threshold_voltage_full and self.current < self.steady_current:
+                self.relative_remaining_capacity = 100 # % of full capacity
+
+            return self.relative_remaining_capacity
+
         else:
             rospy.logwarn("cannot calculate relative remaining capacity. full_charge_capacity=%d, remaining_capacity=%d" % (self.full_charge_capacity, self.remaining_capacity))
             return 0.0
 
+        return self.relative_remaining_capacity
```
mgruhler, line 91: I guess we will never end up here, so this is unnecessary...

cob_bms_driver/src/power_state_aggregator.py
timhappy, line 80: because the current drawn from the battery is negative. I find it unintuitve to edit a negative steady current in the yaml

cob_bms_driver/src/power_state_aggregator.py
timhappy, line 91: This is the case if the bms driver is not running yet as far as I understand this.

mgruhler, line 91: If so, this will fail as this is member is not initialized.
However, three is a `return` in both the `if` and the `else` branch, so we will not get to this part...

benmaidel, line 91: yep the returns are unnecessary. We should remove them

cob_bms_driver/src/power_state_aggregator.py
```diff
@@ -63,11 +71,26 @@ def calculate_power_consumption(self):
 
     def calculate_relative_remaining_capacity(self):
         if self.full_charge_capacity != None and self.remaining_capacity != None:
-            return round(100.0*(self.remaining_capacity/self.full_charge_capacity), 3)
```
floweisshardt, line 66: please keep this line as a comment because this is the right calculation as soon as the bms delivers correct data

benmaidel:
I think that we can close this PR. For now there is this PR https://github.com/ipa320/cob_driver/pull/279. As soon as the bms is working correctly, we can switch back using the power_state_aggregator.

floweisshardt:
this PR is not relevant any more because we're using the phidget based power state publisher which calculates a relative_remaining_capacity

PR #279, opened 2016-07-12
powerstate from phidget

benmaidel:
This is a workaround as long as the bms system is not working!

### DIFF
--- a/cob_phidget_power_state/package.xml
+++ b/cob_phidget_power_state/package.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>cob_phidget_power_state</name>
+  <version>0.0.0</version>
+  <description>The cob_phidget_power_state package</description>
+
+  <maintainer email="bnm@ipa.fhg.de">Benjamin Maidel</maintainer>
+  <autor email="bnm@ipa.fhg.de">Benjamin Maidel</autor>
+
+  <license>BSD</license>
+
+  <buildtool_depend>catkin</buildtool_depend>
+  <exec_depend>cob_msgs</exec_depend>
+  <exec_depend>cob_phidgets</exec_depend>
+  <exec_depend>rospy</exec_depend>
+
+</package>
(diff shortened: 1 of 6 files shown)


floweisshardt:
as this is a cob4-1/msh specific and temporary fix we should NOT merge this into cob_driver (as it would add a dependency from cob_bms_driver to cob_phidgets)

please move script to msh repository

benmaidel:
but the bms bug is a general bug! It does not only concern the msh project. cob4 is not working with the bms driver... Until the bms bug is fixed, this should be the workaround for the cob4.

benmaidel:
i know, having an dependency to another package isn't that nice. But they are inside the same repo...

benmaidel:
as discussed with @ipa-fmw I've moved the phidget power_state script to an extra package

cob_phidget_power_state/package.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>cob_phidget_power_state</name>
+  <version>0.0.0</version>
+  <description>The cob_phidget_power_state package</description>
+
+  <maintainer email="bnm@ipa.fhg.de">Benjamin Maidel</maintainer>
+  <autor email="bnm@ipa.fhg.de">Benjamin Maidel</autor>
+
+  <license>BSD</license>
```
floweisshardt: Please use LGPL

cob_phidget_power_state/package.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>cob_phidget_power_state</name>
+  <version>0.0.0</version>
```
floweisshardt: should be the same version as all other packages in cob_driver (0.6.7)

floweisshardt:
merge together with https://github.com/ipa320/cob_robots/pull/496

cob_phidget_power_state/package.xml
benmaidel: why is this? The `cob_phidget_power_state` pkg is brand new

floweisshardt: the release process unfortunatelly requires all packages within a repository to have the same version number :-(

floweisshardt:
CR: OK

PR #282, opened 2016-07-20
updates for changed hardware

floweisshardt:
phidget hardware changed: now measuring directly at the battery I/O


floweisshardt:
tested on cob4-2: OK

PR #285, opened 2016-08-03
Fix: vlc 2.2 version use by default the wrong video output

NadiaHG:


floweisshardt:
CR: OK

floweisshardt:
tested on cob4-5: OK

PR #286, opened 2016-08-03
BMS: restart CAN interface on failure

mathias-luedtke:
Might take up to 5 seconds / errors msgs to recover

### DIFF
--- a/cob_bms_driver/src/cob_bms_driver_node.cpp
+++ b/cob_bms_driver/src/cob_bms_driver_node.cpp
@@ -460,6 +460,11 @@ void CobBmsDriverNode::diagnosticsTimerCallback(const ros::TimerEvent& event)
 {
 	//update diagnostics
 	updater_.update();
+	if(!socketcan_interface_.getState().isReady()){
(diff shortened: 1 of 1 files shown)


floweisshardt:
this should be tested on cob4-2 as soon as it is back from Ingolstadt and CAN is connected again

mathias-luedtke:
it was tested on cob4-6 and cob4-?

Needs:

```
allow-hotplug can1
iface can1 can static
    bitrate 500000
    pre-up ip link set $IFACE type can restart-ms 10
```

in `/etc/network/interfaces`

floweisshardt:
@ipa-nhg: can you please update the installation docs to include the changes needed for `/etc/network/interfaces`

NadiaHG:
@ipa-fmw it is already documented : https://github.com/ipa320/setup_cob4/blob/master/InstallCob4.sh#L105  and updated https://github.com/ipa320/setup_cob4/pull/33

floweisshardt:
perfect!
CR: OK

PR #288, opened 2016-08-05
new package to calculate em state from phidgets digital input

benmaidel:
depends on: https://github.com/ipa320/cob_robots/pull/515

### DIFF
--- a/cob_phidget_em_state/scripts/em_state_phidget.py
+++ b/cob_phidget_em_state/scripts/em_state_phidget.py
@@ -40,20 +40,20 @@ class EMStatePhidget():
                 got_message = True
 
         if got_message:
-            if (front_em_active && rear_em_active) && (!last_front_em_state && !last_rear_em_state):
+            if (front_em_active and rear_em_active) and (not last_front_em_state and  not last_rear_em_state):
                 self.em_msg.emergency_button_stop = True
                 self.em_caused_by_button = True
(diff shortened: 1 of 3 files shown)


floweisshardt:
FR: OK

cob_phidget_em_state/scripts/em_state_phidget.py
```diff
@@ -0,0 +1,91 @@
...
+        self.sub_digital_sensors = rospy.Subscriber("digital_sensors", DigitalSensor, self.phidget_cb)
+        self.pub_em_state = rospy.Publisher('em_stop_state', EmergencyStopState, queue_size=1)
+
+    def phidget_cb(self, msg):
+        self.last_update = rospy.Time.now()
+        front_em_active = False
+        rear_em_active = False
+        got_message = False
+
+        for i in range(0, len(msg.uri)):
```
floweisshardt, line 34: this seems to be a raw specific mapping, isn't it? can this be outsourced to a yaml file?

benmaidel, line 34: sure, could be outsourced to yaml. But not really necessary for now, because only the raw's will use this script.

benmaidel, line 34: oh and just one remark on writing code comment. Please set your comment below the code block and not above. This way the code lines are also visible on the conversation page.

PR #289, opened 2016-08-17
power_state phidget: fix current sign

floweisshardt:


floweisshardt:
close because PR is splitted into #290 and #291

PR #290, opened 2016-08-19
return true for stop mode service to avoid ugly error messages on client side

floweisshardt:


floweisshardt:
tested on cob4-2: OK

benmaidel:
CR: ok

PR #291, opened 2016-08-19
Feature/phidget power state: 	fix current sign

floweisshardt:


benmaidel:
we should check if this is the same on raw's, otherwise should add an parameter to inverse the current output

floweisshardt:
@ipa-bnm: can you please check for the raws?

benmaidel:
seems to be ok on the raw's. If not, we should correct the cabling...
Will merge this for now

PR #292, opened 2016-09-01
cob_sick_lms1xx: add range configuration

Shin-nn:
Configuration for range added

### DIFF
--- a/cob_sick_lms1xx/ros/src/lms1xx_node.cpp
+++ b/cob_sick_lms1xx/ros/src/lms1xx_node.cpp
@@ -53,6 +53,8 @@ private:
     double resolution;
     double frequency;
     bool set_config;
+    double min_range;
+    double max_range;
 };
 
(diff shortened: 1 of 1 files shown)


destogl:
Are you filter somewhere data based on this ranges?

Shin-nn:
Yeah, we are changing lidars and so on and we need to keep correct ranges.

destogl:
Ah OK, you are doing it in the node which reads data from sensor... Should it be added here for this laser scanner?

Shin-nn:
Given the fact that it is constructional parameter of lidar, there is no other place to set this. It is definitely not correct to set it to 20. If this value is nonsense, you can delete such information from messages, right?

destogl:
Yes, you are right

floweisshardt:
looks good and does not change default behaviour.
CR: OK

benmaidel:
+1

PR #296, opened 2016-09-28
fade volume on play and between tracks

benmaidel:
depends on https://github.com/ipa320/cob_robots/pull/540


floweisshardt (approved):
tested on cob4-1: OK

PR #297, opened 2016-11-23
[kinetic] migration

benmaidel:


mgruhler:
`sound_play` should be released by now into kinetic: http://repositories.ros.org/status_page/ros_kinetic_default.html?q=sound_play

benmaidel:
sound_play is not used by cob_sound!

mgruhler:
whoops, misunderstood the "non exisiting" then ;-) Actually wondered why you would "fix" something like this :-D

mgruhler (approved):

mgruhler:
maybe this should be a squash merge... the merge commits are unnecessary...

NadiaHG (approved):

mathias-luedtke:
* Why do we keep an ipa320 fork of `industrial_ci`? (It has already diverged, upstream will be heavily updated in the near future..)
* if `cob_sound` does not need to depend on `sound_play` this should be fixed for `indigo` as well. ~~@ipa-nhg: why did you add it in #174?~~ (`cob_sound` is based on vlc now ..)
* if we wait 10 more days, `ros_canopen` for kinetic should be synced ;)

benmaidel:
@ipa-mdl 
- we keep an ipa320 fork of `industrial_ci` because the ros_industrial version is producing way to much logging output. On bigger repositories travis will always fail because buffer size exceeds...
- right, i've switched from sound_play to vlc
- nice ;)

mathias-luedtke:
> (@ipa-bnm) we keep an ipa320 fork of industrial_ci because the ros_industrial version is producing way to much logging output. On bigger repositories travis will always fail because buffer size exceeds...

So at least keep it sync..
It would even better if we could provide a patch to upstream with a flag to enter silent mode
I guess the build will be much more silent after the refactoring.

benmaidel:
>So at least keep it sync..
It would even better if we could provide a patch to upstream with a flag to enter silent mode
I guess the build will be much more silent after the refactoring.

My solution to keep it more silent was already discussed in one of the issues and was not accepted. So I've decided to provide no patch. If, in the Future, a better no silent version is available we could still switch to ros_industrial version!?
But i will try to keep it more sync with upstream.

PR #307, opened 2017-02-02
[HOTFIX] fix powerstate aggregator charging flag

floweisshardt:
bms is not delivering correct flag for full battery and docked

### DIFF
--- a/cob_bms_driver/src/power_state_aggregator.py
+++ b/cob_bms_driver/src/power_state_aggregator.py
@@ -38,7 +38,7 @@ class PowerStateAggregator():
             self.last_currents.pop(0)
         self.last_currents.append(msg.data)
 
-        if msg.data > -1: # we use a limit of -1Ampere because if the battery is 100% fuland the robot is still docked, there is no more current going into the battery. -1 A is biggger than the "Ruhestrom", so this should be ok until BMS is fixed and delivers a proper flag for docked or not_docked.
+        if msg.data > -1: # we use a limit of -1Ampere because if the battery is 100% full and the robot is still docked, there is no more current going into the battery. -1 A is biggger than the "Ruhestrom", so this should be ok until BMS is fixed and delivers a proper flag for docked or not_docked.
             self.charging = True
         else:
             self.charging = False
--- a/cob_bms_driver/src/power_state_aggregator.py
+++ b/cob_bms_driver/src/power_state_aggregator.py
@@ -2,7 +2,7 @@
 
 import rospy
 import numpy
-from std_msgs.msg import Bool, Float64
+from std_msgs.msg import Float64
 from cob_msgs.msg import PowerState
 
 class PowerStateAggregator():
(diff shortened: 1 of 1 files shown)


cob_bms_driver/src/power_state_aggregator.py
```diff
@@ -51,10 +55,6 @@ def temperature_cb(self, msg):
         self.last_update = rospy.Time.now()
         self.temperature = msg.data
 
-    def charging_cb(self, msg):
-        self.last_update = rospy.Time.now()
-        self.charging = msg.data
```
fmessmer, line 56: So, if this is a temporary workaround and the flag will later be retrieved from bms again, don't you want to keep the subscriber - simply comment `self.charging = msg.data`?

cob_bms_driver/src/power_state_aggregator.py
```diff
@@ -36,8 +35,13 @@ def current_cb(self, msg):
 
         # fill current into list of past currents for filtering purposes
         if len(self.last_currents) >= self.current_buffer_size:
-            self.last_currents.pop()
+            self.last_currents.pop(0)
         self.last_currents.append(msg.data)
+
+        if msg.data > -1: # we use a limit of -1Ampere because if the battery is 100% fuland the robot is still docked, there is no more current going into the battery. -1 A is biggger than the "Ruhestrom", so this should be ok until BMS is fixed and delivers a proper flag for docked or not_docked.
```
fmessmer: ....`fuland` :arrow_forward: `full and`....

fmessmer (commented):
Review

ipa-fxm:
@ipa-fmw 
Even better, could we not apply this workaround in bms driver itself? I.e. set the `charging` flag using the metric you propose, i.e. `current < -1 A`? That way we would not need to "remember" adjusting the power_state_aggregator after fixing bms...!

fmessmer (changes requested):
@ipa-fmw 
Even better, could we not apply this workaround in bms driver itself? I.e. set the `charging` flag using the metric you propose, i.e. `current < -1 A`? That way we would not need to "remember" adjusting the power_state_aggregator after fixing bms...!

benmaidel:
Isn't there any way to detect if the robot is connected to the docking station? This way we could introduce a new flag `docked` and leave the charging flag as it is!

cob_bms_driver/src/power_state_aggregator.py
floweisshardt: done

floweisshardt:
@ipa-fxm: we cannot fix this in the bms driver because the driver only publishes raw data. the power_state_aggregator is the node which accumulated the various topics from the bms driver into a consistent power_state

@ipa-bnm: no there is no direct way to detect the `docked` status. What we can do is to keep charging as it is (current switch = 0) and introduce a new field docked in power_state (current switch = -1). But changing the PowerState message this will lead to inconsistent and not analysable bag files from Oct to now. So I prefer to keep the workaround and fix it properly as soon as the bms is capable of detecting this (probably takes until summer 20XX)

floweisshardt:
I'd like to merge this as we're finalising testing on cob4-2 and I'd like to ship an up-to-date but well tested installation.

fmessmer:
workaround for #308

Issue #249, opened 2015-11-19
[cob_scan_unifier] cob_scan_unifier is not merging complete scans in simulation


benmaidel:
_From @ipa-bnm on September 17, 2015 12:5_

scan unifier produce gap on the back of cob4-2
![screenshot from 2015-09-17 14 01 09](https://cloud.githubusercontent.com/assets/483721/9932505/bbfbda80-5d44-11e5-9da1-362de7ea780a.png)

@ipa-frm @ipa-mig @fmessmer FYI

_Copied from original issue: ipa320/cob_navigation#53_

benmaidel:
_From @ipa-mig on September 18, 2015 6:1_

hmm, also, there is a z-offset (which will probably not be a problem for our use-cases... but we should change that anyways...

benmaidel:
the z-offset seems to be ok. Imagine the scanners are mounted in different z positions (eg. our BMW). So transform the scans to the ground is ok, from my opinion. A better solution would be the usage of a point cloud instead a merged scan message.

benmaidel:
https://github.com/iralabdisco/ira_laser_tools 
there you can find a tool with the same purpose as cob_scan_unifier.
Difference: 
- no gaps
- dynamic reconfigure
- offers point cloud 2 as well

Perhaps we should stick to this package? We could even copy the code from this package to our repo, because it has no license at all.

destogl:
@ipa-bnm yesterday I ported cob_scan_unifier to tf2 (s. #257), are you proposing to use this other repository and let cob_scan_unifier to die out? This is important to me to know in which direction to go.

Regarding this not merging scans, I also saw it with real scans but can not really say where is the problem.

benmaidel:
@destogl for now we will stick to cob_scan_unifier. For the moment there are no plans to switch to another solution. But if we continue to have problems with cob_scan_unifier , it could be that we change or reimplement it. I will also have a look at your PR

mgruhler:
Also, check the angle increment, see [comment](https://github.com/ipa320/cob_driver/pull/324#discussion_r105103344)

fmessmer:
@ipa-mig, is @mig-em working on this after we merged #324?

mgruhler:
he should be :-)

ghost:
I can't reproduce this bug in the current scan_unifier. I tested by rotating in a corner of ipa-apartment and farer away to increase the effect, but no gaps showed up.

fmessmer:
[...]

Issue #248, opened 2015-11-19
[cob_scan_unifier] Update rate


fmessmer:
Some discussion history from https://github.com/ipa320/cob_robots/pull/367#issuecomment-157902641

> @ipa-fxm:
> On a side note:
> @ipa-fmw thought that 100Hz might be faster than the actual scanner publish rate....Should the rate be > reduced?
> @ipa-bnm What do you think?
> 
> @ipa-mig:
> @ipa-fxm @ipa-fmw @ipa-bnm Yes, the rate of 100Hz is (considerably) faster than the scanner publish rate (which is 12Hz for the S300, afaik). However, this is on purpose to send out the combined scan asap.
> It is not deterministic in which order the single scan topics come in and it is, in the worst case, almost one full scan rate cycle delay. This is due to the scan_unifier waiting until it received information from every scan before it combines one.
> This frequency was chosen to be high to check the subscriber callbacks frequently to introduce as little latency between the original scans and the combined one as possible.
> However, it could still be beneficial to change this frequency...

Some hints that might help reduce the update rate while guaranteeing the desired behaviour:
- In order to ensure a certain publish rate for the assembled scan, a [ros::Timer](http://wiki.ros.org/roscpp/Overview/Timers) callback can/should be used
- in order to sync multiple input topic (i.e. the scan topics) depending on their arrival time, [message_filters](http://wiki.ros.org/message_filters) can/should be used (e.g. [TimeSynchronizer](http://wiki.ros.org/message_filters#Time_Synchronizer))

Some questions to decide on best implementations:
- What is more important: a guaranteed publish rate for the assembled scan? or that the assembled scan is complete, i.e. it consists of the latest scan from each of the configured input scans?
- Would it be acceptable to publish a partial assembled scan, i.e. do not include scans from **all** input scans in case one did not arrive within a given time period?

---

@ipa-bnm could you also create a new issue for the assembling bug that you told me about? so that it is documented!

mgruhler:
@ipa-bnm btw, you told me once about another scan unifier. Where is this again?

@ipa-fxm for some mapping algorithms, it is important to have a unified scan of _all_ scans. The problem is actually not in the algorithm itself, but in the, sometimes very, bad calibration of the laser scanners. this results in a "rotational jumping" which produces bad maps. this is why this has been developed in the first place.

For localization, I don't need this at all.

Thus, it is actually hard to say, what would be okay. I cannot really tell ;-)

Where else is the scan_unifier required currently?

[...]

PR #313, opened 2017-02-08
Skip laserscans with no corresponding tf transform

### LINKED ISSUE #268: [cob_scan_unifier] LookupError on start up
Appears once on startup, probably simply requires a sleep after tf::TransformListener initialization (for filling the buffer):

```
Node: /scan_unifier
Time: 16:05:39.612337463 (2016-03-04)
Severity: Error
Published Topics: /rosout, /scan_unified

Lookup would require extrapolation into the past.  Requested time 1457103936.557201524 but the earliest data is at time 1457103937.074180613, when looking up transform from frame [base_laser_front_link] to frame [base_link]

Location:
/u/fxm/git/care-o-bot/src/cob_driver/cob_scan_unifier/src/scan_unifier_node.cpp:LaserScan scan_unifier_node::unifieLaserScans:252
```

Also, `unifieLaserScans` should be `unifyLaserScans` :wink:

ghost:
fixes https://github.com/ipa320/cob_driver/issues/268

### DIFF
--- a/cob_scan_unifier/src/scan_unifier_node.cpp
+++ b/cob_scan_unifier/src/scan_unifier_node.cpp
@@ -123,6 +123,16 @@ void scan_unifier_node::getParams()
     config_.number_input_scans = 0;
     ROS_ERROR("No parameter input_scans on parameter server!! Scan unifier can not subscribe to any scan topic!");
   }
+
+  if(!pnh_.hasParam("frame"))
+  {
+    ROS_WARN("No parameter frame on parameter server. Using default value [base_link].");
+    frame_ = "base_link";
+  }
+  else
+  {
+    pnh_.getParam("frame", frame_);
+  }
 }
 
 /**
@@ -242,13 +252,13 @@ sensor_msgs::LaserScan scan_unifier_node::unifyLaserScans()
       ROS_DEBUG_STREAM("Converting scans to point clouds at index: " << i << ", at time: " << vec_laser_struct_.at(i).current_scan_msg.header.stamp << " now: " << ros::Time::now());
       try
       {
-        if (!listener_.waitForTransform("base_link", vec_laser_struct_.at(i).current_scan_msg.header.frame_id,
+        if (!listener_.waitForTransform(frame_, vec_laser_struct_.at(i).current_scan_msg.header.frame_id,
             vec_laser_struct_.at(i).current_scan_msg.header.stamp, ros::Duration(3.0)))
             {
               continue;
             }
         ROS_DEBUG("now project to point_cloud");
-        projector_.transformLaserScanToPointCloud("base_link",vec_laser_struct_.at(i).current_scan_msg, vec_cloud.at(i), listener_);
+        projector_.transformLaserScanToPointCloud(frame_,vec_laser_struct_.at(i).current_scan_msg, vec_cloud.at(i), listener_);
       }
       catch(tf::TransformException ex){
         ROS_ERROR("%s",ex.what());
@@ -256,7 +266,7 @@ sensor_msgs::LaserScan scan_unifier_node::unifyLaserScans()
     }
     ROS_DEBUG("Creating message header");
     unified_scan.header = vec_laser_struct_.at(0).current_scan_msg.header;
-    unified_scan.header.frame_id = "base_link";
+    unified_scan.header.frame_id = frame_;
     unified_scan.angle_increment = M_PI/180.0/2.0;
     unified_scan.angle_min = -M_PI + unified_scan.angle_increment*0.01;
     unified_scan.angle_max =  M_PI - unified_scan.angle_increment*0.01;
@@ -350,6 +360,9 @@ int main(int argc, char** argv)
   ros::init(argc, argv, "cob_scan_unifier_node");
 
   scan_unifier_node my_scan_unifier_node;
+
+  //Sleep to make sure tf_listener buffer fills up
+  ros::Duration(1).sleep();
 
   // store initialization time of the node
   ros::Time start = ros::Time::now();
--- a/cob_scan_unifier/src/scan_unifier_node.cpp
+++ b/cob_scan_unifier/src/scan_unifier_node.cpp
@@ -220,14 +220,14 @@ void scan_unifier_node::topicCallbackLaserScan(const sensor_msgs::LaserScan::Con
 }
 
 /**
- * @function unifieLaserScans
- * @brief unifie the scan information from all laser scans in vec_laser_struct_
+ * @function unifyLaserScans
+ * @brief unify the scan information from all laser scans in vec_laser_struct_
  *
  * input: -
  * output:
  * @param: a laser scan message containing unified information from all scanners
  */
-sensor_msgs::LaserScan scan_unifier_node::unifieLaserScans()
+sensor_msgs::LaserScan scan_unifier_node::unifyLaserScans()
 {
   sensor_msgs::LaserScan unified_scan = sensor_msgs::LaserScan();
   std::vector<sensor_msgs::PointCloud> vec_cloud;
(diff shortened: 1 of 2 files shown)


cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -242,11 +242,13 @@ sensor_msgs::LaserScan scan_unifier_node::unifieLaserScans()
...
-            vec_laser_struct_.at(i).current_scan_msg.header.stamp, ros::Duration(3.0));
-
+        if (!listener_.waitForTransform("base_link", vec_laser_struct_.at(i).current_scan_msg.header.frame_id,
+            vec_laser_struct_.at(i).current_scan_msg.header.stamp, ros::Duration(3.0)))
+            {
+              continue;
+            }
         ROS_DEBUG("now project to point_cloud");
-        projector_.transformLaserScanToPointCloud("/base_link",vec_laser_struct_.at(i).current_scan_msg, vec_cloud.at(i), listener_);
+        projector_.transformLaserScanToPointCloud("base_link",vec_laser_struct_.at(i).current_scan_msg, vec_cloud.at(i), listener_);
```
mgruhler: could you please make this a parameter, while you are at it?

mgruhler (changes requested):
Could you also add the sleep (1s) after the initialization of the transform listener, as proposed in the issue.

Also, please investigate (in simulation) what happens if you set the frequency of this node to something higher than 10 Hz.

ghost:
Frequency was already 100Hz. Tried also with 15Hz and it works. Should I add the param to the YAMLs of all robots?

mgruhler:
@mig-em yes, please do that.

fmessmer:
could we also investigate #248 and #249 while @mig-em is on it?
(At least check and report whether the issues are still an issue :wink:)

mgruhler:
@mig-em Please do as commanded ;-)

mgruhler (approved):

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -242,19 +252,21 @@ sensor_msgs::LaserScan scan_unifier_node::unifieLaserScans()
...
       try
       {
-        listener_.waitForTransform("/base_link", vec_laser_struct_.at(i).current_scan_msg.header.frame_id,
-            vec_laser_struct_.at(i).current_scan_msg.header.stamp, ros::Duration(3.0));
-
+        if (!listener_.waitForTransform(frame_, vec_laser_struct_.at(i).current_scan_msg.header.frame_id,
+            vec_laser_struct_.at(i).current_scan_msg.header.stamp, ros::Duration(3.0)))
+            {
+              continue;
+            }
```
benmaidel: please correct indentation

benmaidel (changes requested):

benmaidel:
@mig-em @ipa-mig isn't 15Hz a bad idea? See your comments here: https://github.com/ipa320/cob_robots/pull/367

I think we should switch to https://github.com/iralabdisco/ira_laser_tools because they seem to have a better implementation. I've already used their implementation once and it worked quite well.

mgruhler:
@ipa-bnm to avoid misunderstandings: @ipa-nhg reported in #268 some problems with a rate higher then 10Hz. This should not be made the default :-)

The comment about the parameter is only to add the new parameter to the yamls...

ghost:
So should I adapt the synchronisation part by using the TimeSynchronizer filter and investigate https://github.com/ipa320/cob_driver/issues/249 or better use the ira_laser_tools?
If I understand it right the rate param will become obsolete by using the filter.

fmessmer:
I'd say we should evaluate `ira_laser_tools` and - if it solves all our problems - use it!
If not, fix issues in current `scan_unifier`...

fmessmer:
Well, if `ira_laser_tools` works for us, we might want to "take over" the package and copy/move it to an ipa320 repo...as it is not released...License is [TODO](https://github.com/iralabdisco/ira_laser_tools/blob/master/package.xml#L9) :wink:

benmaidel (approved):

mgruhler:
if `ira_laser_tools` is actually working good, "taking it over" would be a good idea. But I'd say we try to get in contact with the developers and see how to proceed. Last commit was Oct. 2015 for that matter...

mgruhler:
@ipa-bnm would this be worth investigating on cob4?

benmaidel:
@ipa-mig yes, we can/should test this on cob4 as soon as a new hardware is available

fmessmer:
Well, would #257 have solved #268 already?

I guess we should keep #248 and #249 out of this PR...as it is separate and needs more testing.
But, I'd like to finalize/accept/close #313, #257 and #268...
@mig-em @ipa-mig @destogl

ghost:
I tried the ira_laser_tools and they seem to work but https://github.com/ipa320/cob_driver/pull/257 and https://github.com/ipa320/cob_driver/issues/268 are not fixed (there is still the "Lookup would require extrapolation into the past." error and they use tf1).
ira_laser_merger

In this screen you see raw3-3 scans (red, green) and the unified scan (yellow)

fmessmer:
Is the following summary correct?
@ipa-mig @mig-em @ipa-bnm 

`cob_scan_unifier`:
 - no LookupTransform error #313 #268 
 - tf2 (tested?) #257 
 - has issues merging scans #249 (were you able to reproduce #249 with your tf-fixes?)

`ira_laser_tools`
 - has LookupTransform error
 - uses tf1
 - merges scans correctly (both sim and real sensor?)
 - release issue as we are not maintainer...

benmaidel:
Just for completeness.
There is also the possibility to use a combination of https://github.com/ros-perception/laser_assembler and https://github.com/ros-perception/pointcloud_to_laserscan to achieve a merged scan.

ghost:
Additionally the parameters of the ira_laser_tools should be adapted.
Maybe the LookupTransform error should be fixed in the driver of the scanners, or is there a point in sending out scans without according transforms to base_link?

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -242,19 +252,21 @@ sensor_msgs::LaserScan scan_unifier_node::unifieLaserScans()
       ROS_DEBUG_STREAM("Converting scans to point clouds at index: " << i << ", at time: " << vec_laser_struct_.at(i).current_scan_msg.header.stamp << " now: " << ros::Time::now());
       try
       {
-        listener_.waitForTransform("/base_link", vec_laser_struct_.at(i).current_scan_msg.header.frame_id,
-            vec_laser_struct_.at(i).current_scan_msg.header.stamp, ros::Duration(3.0));
-
+        if (!listener_.waitForTransform(frame_, vec_laser_struct_.at(i).current_scan_msg.header.frame_id,
+            vec_laser_struct_.at(i).current_scan_msg.header.stamp, ros::Duration(3.0)))
+        {
+          continue;
```
fmessmer, line 258: @ipa-mig @mig-em 
Well, here we silently skip a scan that cannot be transformed. Should we at least notify in some way?
e.g. `ROS_ERROR_STREAM("Not using scan during unification: " << vec_laser_struct_.at(i).current_scan_msg.header.frame_id,);`

fmessmer (changes requested):

fmessmer:
@ipa-mig FYI

As discussed with @mig-em, we will try a [message_filter](http://wiki.ros.org/message_filters)-based implementation before merging this PR

mgruhler:
good idea.

fmessmer:
message filter implementation is in #324 
In case #324 works, this PR is obsolete

fmessmer:
https://github.com/ipa320/cob_driver/pull/324 is the proper way to solve #268

PR #317, opened 2017-02-13
[cob_sick_s300] fix bug reading not all scan points

### LINKED ISSUE #256: [cob_sick_s300] missing distance values of sick s300 driver
@ipa-josh  
@ipa-mig 

When using the published scan topics of "cob_sick_s300.cpp" in each message there are 5 distance values (10 Byte) missing.
The above mentioned class is based on the sick S300 driver (e. g. "TelegramS300.h") that reads the telegrams sent by the S300.
One part of this telegram contains the distance values. In "TelegramS300.h" (size_t num_points) the length of these values inside the telegram is calculated.
https://github.com/ipa320/cob_driver/blob/95e4bf0525cb13630fa768fd6394357d9423052f/cob_sick_s300/common/include/cob_sick_s300/TelegramS300.h#L313-L327

Note: Values at the end of each line in Byte except the tc1_.size in 16 bit words.
                                                                              should be 1082!
//std::cout << "num_points: " << num_points << std::endl; // 1072
/*std::cout << "--------------- start ------------------" << std::endl
          << "tc1_.size: " << tc1_.size << std::endl                        // 552
          << "sizeof(tc1_): " << sizeof(tc1_) << std::endl                  // 10
          << "sizeof(tc2_): " << sizeof(tc2_) << std::endl                  // 10
          << "sizeof(tc3_): " << sizeof(tc3_) << std::endl                  // 2
          << "sizeof(td_): " << sizeof(td_) << std::endl                    // 2
          << "sizeof(TELEGRAM_TAIL): " << sizeof(TELEGRAM_TAIL) << std::endl// 2
          << "JUNK_SIZE: " << JUNK_SIZE << std::endl                        // 4
          << "last_offset_: " << last_offset_ << std::endl                  // -10
          << "--------------- end ------------------" << std::endl;
*/

Telegram example:

| name | example value | size in Byte |
| --- | --- | --- |
| reply_telegram: | 0000 0000 | 4 = JUNK_SIZE |
| trigger_result: | 0000 | 2 |
| reply telegram does not count to size!!! |  |  |
| size (16 bit words): | 0228 = 552 dec | 2 |
| coordination_flag: | ff | 1 |
| device_addresss: | 07 | 1 |
| sum |  | 10 = tc1_ |
| protocol_version: | 0201 | 2 |
| status: | 0000 | 2 |
| scan_number: | 6001 0000 | 4 |
| telegram_number: | 5800 | 2 |
| sum |  | 10 = tc2_ |
| type (ID for data): | bbbb | 2 |
| sum |  | 2 = tc3_ |
| type (ID for segments in data): | 1111 | 2 |
| sum |  | 2 = td_ |
| measurement data: | 1300 .. 1c00 | 1082 |
| crc: | ffdf | 2 = TELEGRAM_TAIL |

=> "last_offset" (-10 Byte) is not needed
10 Byte data is missing!

I could not figure out why last_offset is set to -10 when telegram version is 0102.
https://github.com/ipa320/cob_driver/blob/95e4bf0525cb13630fa768fd6394357d9423052f/cob_sick_s300/common/include/cob_sick_s300/TelegramS300.h#L243-L244

S3000/S300 Expert telegram listing documentation does not tell anything about an offset or something similar.
[S3000_S300_Expert_2.10_telegram_listing.pdf](https://github.com/ipa320/cob_driver/files/69662/S3000_S300_Expert_2.10_telegram_listing.pdf)

Testing with "last_offset" set to 0 (hardcoded) solves this issue, the scan topic now publishes all the data in full length with reasonable values.

### LINKED ISSUE #276: Lidar scan is warped
I had some issues with two sick S300 Professional on my robot and could not get the scans to overlap even after calibrating the lidars realtive to robot center. As it was not caused by some offsets in the lidars pose, I started checking the what the lidar driver was delivering. I also found another [S300 driver](https://github.com/bohlender/sicks300) that delivered a slightly different scan while observing the same thing. This motivated me to look more into this problem. I found out that the scans I get using your driver have an angle increment of slightly more than 0.5 degrees and is computed from angle_start, angle_stop and size of the scan array, while the angle increment in the other driver I tested was explicitly set to 0.5, number that is stated in the data sheet of the Sick S300.
In order to get my IPA Sick driver to deliver a similar signal to the other driver, I had to change [this file](https://github.com/ipa320/cob_driver/blob/indigo_dev/cob_sick_s300/common/src/ScannerSickS300.cpp) and set the "- 1" in line 264 to "+ 4". I am still not sure about the reason why it works this way but now the increment computed is equal to 0.5 degrees and the extrinsic calibration works and the front and rear laser are overlapping. Can you please take a look into this?

mgruhler:
This PR fixes #256 and fixes #276 by (more) correctly implementing the size calculations acoording to the (updated) Sick telegram listing.

Tested with:
- [x] S300 Pro CMS (old protocol)
- [x] S300 Expert (compatability mode)
- [x] S300 Expert (new protocol)

This PR also comments out a warning about "old" parameter sets, as those parameter sets actually are buggy --> #318

@ipa-josh

### DIFF
--- a/cob_sick_s300/README.md
+++ b/cob_sick_s300/README.md
@@ -4,7 +4,7 @@ It provides an implementation for both, the old (1.40) and the new (2.10) protoc
 Thus, the old Sick S300 Professional CMS as well as the new Sick S300 Expert are supported.
 
 However, it does not cover the full functionality of the protocol:
-- It only handle distance measurements properly
+- It only handles distance measurements properly
 - It only handles no or only one configured measurement range field properly
 - It does not handle I/O-data or reflector data
 (though it reads the reflector marker field in the distance measurements)
--- a/cob_sick_s300/README.md
+++ b/cob_sick_s300/README.md
@@ -1,0 +1,22 @@
+# cob_sick_s300
+This package implements a driver for the Sick S300 Safety laser scanners.
+It provides an implementation for both, the old (1.40) and the new (2.10) protocol.
+Thus, the old Sick S300 Professional CMS as well as the new Sick S300 Expert are supported.
+
+However, it does not cover the full functionality of the protocol:
+- It only handle distance measurements properly
+- It only handles no or only one configured measurement range field properly
+- It does not handle I/O-data or reflector data
+(though it reads the reflector marker field in the distance measurements)
+
+See http://wiki.ros.org/cob_sick_s300 for more details.
+
+## S300 Configuration
+Here are a few notes about how to best configure the S300:
+- Configure the RS422 output to 500kBaud (otherwise, the scanner only provides a lower frequency)
+- Configure the scanner to Continuous Data Output
+- Send data via one telegram
+- Only configure distances, no I/O or reflector data (otherwise, the scanner only provides a lower frequency).
+- Do not configure a measurement range field (otherwise, the scanner only provides a lower frequency).
+If you want to only use certain measurement ranges, do this on the ROS side using e.g. the `cob_scan_filter`
+located in this package as well.
--- a/cob_sick_s300/README.md
+++ b/cob_sick_s300/README.md
@@ -17,6 +17,9 @@ Here are a few notes about how to best configure the S300:
 - Configure the scanner to Continuous Data Output
 - Send data via one telegram
 - Only configure distances, no I/O or reflector data (otherwise, the scanner only provides a lower frequency).
-- Do not configure a measurement range field (otherwise, the scanner only provides a lower frequency).
-If you want to only use certain measurement ranges, do this on the ROS side using e.g. the `cob_scan_filter`
+- Configuration of the measurement ranges
+    - For protocol 1.40: only configure one measurement range field with the full range (-45° to 225°) with all values.
+    - For protocol 2.10: do not configure a measurement range field
+      (otherwise, the scanner only provides a lower frequency).
+- If you want to only use certain measurement ranges, do this on the ROS side using e.g. the `cob_scan_filter`
 located in this package as well.
--- a/cob_sick_s300/common/include/cob_sick_s300/TelegramS300.h
+++ b/cob_sick_s300/common/include/cob_sick_s300/TelegramS300.h
@@ -62,7 +62,7 @@ class TelegramParser {
 		struct {
 			uint32_t reply_telegram;
 			uint16_t trigger_result;
-			uint16_t size;
+			uint16_t size; // in 16bit=2byte words
 			uint8_t  coordination_flag;
 			uint8_t  device_addresss;
 		};
@@ -211,10 +211,14 @@ class TelegramParser {
 	TELEGRAM_COMMON2 tc2_;
 	TELEGRAM_COMMON3 tc3_;
 	TELEGRAM_DISTANCE td_;
-	int last_offset_;
+	int length_count_start_, length_count_crc_, user_data_size_;
 public:
 
-	TelegramParser() : last_offset_(0) {}
+	TelegramParser() :
+		length_count_start_(0),
+		length_count_crc_(0),
+		user_data_size_(0)
+	{}
 
 	bool parseHeader(const unsigned char *buffer, const size_t max_size, const uint8_t DEVICE_ADDR, const bool debug)
 	{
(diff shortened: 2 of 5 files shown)


mgruhler:
Testing with S300 Expert and the new protocol revealed some issues. Will reinvestigate this next week.

mgruhler:
@ipa-fxm @ipa-josh please review
@ipa-tif fyi

cob_sick_s300/README.md
```diff
@@ -0,0 +1,25 @@
+# cob_sick_s300
+This package implements a driver for the Sick S300 Safety laser scanners.
+It provides an implementation for both, the old (1.40) and the new (2.10) protocol.
+Thus, the old Sick S300 Professional CMS as well as the new Sick S300 Expert are supported.
+
+However, it does not cover the full functionality of the protocol:
+- It only handle distance measurements properly
```
fmessmer: `handle*s*`

cob_sick_s300/common/include/cob_sick_s300/TelegramS300.h
```diff
@@ -229,47 +255,87 @@ class TelegramParser {
...
+		// The size reported by the protocol varies depending on the calculation which is different depending
+		// on several factors.
+		// The calculation is described on pp. 70-73 in:
+		// https://www.sick.com/media/dox/1/91/891/Telegram_listing_S3000_Expert_Anti_Collision_S300_Expert_de_en_IM0022891.PDF
+		//
+		// Also, the size is reported as 16bit-words = 2 bytes...
+		//
+		// For the old protocol/compatability mode:
+		// "The telegram size is calculated ... starting with the ... 5. byte ... up to and including the ... CRC."
 		if(tc2_.protocol_version==0x102)
```
fmessmer, line 274: would it make sense to have some kind of `if(0x102)-elseif(0xXXX)-else(unsupported-protocol)` construct?

cob_sick_s300/ros/src/cob_sick_s300.cpp
```diff
@@ -188,7 +188,7 @@ class NodeClass
 				}
 				else
 				{
-					ROS_WARN("No params for the Sick S300 fieldset were specified --> will using default, but it's deprecated now, please adjust parameters!!!");
+					//ROS_WARN("No params for the Sick S300 fieldset were specified --> will using default, but it's deprecated now, please adjust parameters!!!");
```
fmessmer, line 191: You know how I don't like commenting warnings...
In case it's just a deprecated parameter syntax, the user should see this warning and then after setting the parameter correctly this warning won't occur anymore...

fmessmer (commented):
Other than the following comments, I can't say much about the protocol parsing...

cob_sick_s300/common/include/cob_sick_s300/TelegramS300.h
mgruhler: Well, probably yes? However, then we could just throw a warning and stop the node. There is no other way to do this. New protocol number is `0x103` (if the documentation is correct at least this once...).

Should I add this?

cob_sick_s300/ros/src/cob_sick_s300.cpp
mgruhler: I do. However, this is a newly introduced parameter which should deprecate the old one. But this is buggy/not properly implemented and undocumented. See issue #318 

The warning pops up every time you use a laserscanner without the new parameter layout and it took me quite some time to get the syntax right. It is also not yet used with cob (at least, I did not find anything).

Thus, instead of telling people they have a deprecated parameter set, we should first make sure this one is working correctly. Also, there is another issue with this parameter as it could leed to ambiguities. I'll gladly discuss this in person...

cob_sick_s300/common/include/cob_sick_s300/TelegramS300.h
fmessmer: Hmm, do we have a chance to "try" the protocols supported and then exit in case it is an unsupported protocol? How would we detect whether the protocol is not supported? Would "total message length" be an indicator?

mgruhler: no, total message length is not an indicator for the protocol. This changes depending on what you configure in the laser scanner itself.

I could throw an exception when there is an unsupported protocol, and "try" in the ROS (or the driver) part. What would you prefer?

mgruhler: Detecting if the protocol is not supported could IMO only be done via checking the protocl number. However, this node could use some more error handling in this parser :-)

However, this would then also be another issue...

mgruhler:
@ipa-fxm @ipa-josh @ipa-fmw ping

ipa-josh (approved):

floweisshardt (approved):
FR: OK

cannot say much about the implemented functionality. I'll open new issues for the comments from @ipa-fxm: https://github.com/ipa320/cob_driver/issues/325

fmessmer:
I'll approve as soon as there are a little more details in #318

fmessmer (approved):

PR #319, opened 2017-02-14
update to industrial_ci

souravran:


fmessmer:
Do we want `industrial_ci` in `indigo_dev`, @ipa-bnm @ipa-mdl @ipa-fmw?
What's the motivation for this PR?

.travis.yml
```diff
@@ -1,23 +1,21 @@
-sudo: required
-dist: trusty
-# Force travis to use its minimal image with default Python settings
+sudo: required 
+dist: trusty 
 language: generic
 compiler:
```
mathias-luedtke: unrelated: setting the compiler is not supported/used if language=generic

.travis.yml
```diff
@@ -1,23 +1,21 @@
...
 install:
-  - wget https://raw.githubusercontent.com/ipa320/care-o-bot/${CI_ROS_DISTRO}_dev/.travis.install.sh
-  - bash .travis.install.sh
-
-script:
-  - wget https://raw.githubusercontent.com/ipa320/care-o-bot/${CI_ROS_DISTRO}_dev/.travis.script.sh
-  - bash .travis.script.sh
+  - git clone https://github.com/ros-industrial/industrial_ci.git .ci_config
+script: 
+  - source .ci_config/travis.sh
```
mathias-luedtke: please do not source, just `.ci_config/travis.sh`

mathias-luedtke:
@ipa-fxm: the idea is to unify the tests for all branches

Do you need the downstream tests to be run?
This currently does not work properly (https://github.com/ros-industrial/industrial_ci/issues/106)
I will address this issue soon.

.travis.yml
floweisshardt: @ipa-bnm: this is copied from the kinetic template. should we consider that for kinetic too?

.travis.yml
floweisshardt: this means we can get rid of compiler parameter at all?

souravran: I think we can either have language specific tags and if the language tag is generic then compiler parameter can be omitted.

floweisshardt (approved):

.travis.yml
benmaidel: yes we should!

PR #323, opened 2017-03-05
Multi distro travis

fmessmer:
tested CI locally: OK

`ipa320/kinetic_dev` branch should be removed after merge!
(see https://github.com/ipa320/cob_driver/compare/indigo_dev...kinetic_dev)

### DIFF
--- a/.travis.rosinstall
+++ b/.travis.rosinstall
@@ -1,8 +1,1 @@
-- git:
-    uri: 'https://github.com/ipa320/cob_common.git'
-    local-name: cob_common
-    version: indigo_dev
-- git:
-    uri: 'https://github.com/ipa320/cob_extern.git'
-    local-name: cob_extern
-    version: indigo_dev
+# empty
--- a/.travis.rosinstall.jade
+++ b/.travis.rosinstall.jade
@@ -1,0 +1,1 @@
+.travis.rosinstall.kinetic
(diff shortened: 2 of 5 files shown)


fmessmer:
@ipa-mdl 
How come Travis-CI is successful here for jade/kinetic with `cob_extern` in the `.travis.rosinstall.jade`/`.travis.rosinstall.kinetic` which obviously fails for all distros in https://github.com/ipa320/cob_extern/pull/79 due to 
 > mkdir: cannot create directory 'build': Read-only file system

mathias-luedtke:
For some reason the `cob_extern` package fail:
```
Makefile:102: *** "Can't find /usr/src/linux-headers-4.4.0-51-generic/include/linux/version.h".  Stop.

Makefile:102: *** "Can't find /usr/src/linux-headers-4.4.0-51-generic/include/linux/version.h".  Stop.
```
This is not handled properly in the wrapping `CMakeLists.txt` and the build continues.

The reason for the failure is that the Makefile uses `uname` to resolve the kernel headers.
This does not work with docker as the kernel is from the host and this particular version is not installed in the image.

mathias-luedtke:
work-around: https://github.com/ipa320/cob_extern/pull/80

(this makes the tests fail again ^^)

NadiaHG (approved):

PR #324, opened 2017-03-08
Fix/rewrite sync

### LINKED ISSUE #268: [cob_scan_unifier] LookupError on start up

ghost:
Use message filters to sync the scans.

fix https://github.com/ipa320/cob_driver/issues/268

### DIFF
--- a/cob_scan_unifier/src/scan_unifier_node.cpp
+++ b/cob_scan_unifier/src/scan_unifier_node.cpp
@@ -109,6 +109,7 @@ ScanUnifierNode::ScanUnifierNode()
       return;
   }
 
+  ros::Duration(1.0).sleep();
 }
 
 // Destructor
@@ -128,27 +129,6 @@ ScanUnifierNode::~ScanUnifierNode()
  */
 void ScanUnifierNode::getParams()
 {
-//  XmlRpc::XmlRpcValue topicList;
-//
-//  // TODO short parsing
-//  if (pnh_.getParam("input_scans", topicList))
-//  {
-//    ROS_ASSERT(topicList.getType() == XmlRpc::XmlRpcValue::TypeArray);
-//
-//    if(topicList.getType() == XmlRpc::XmlRpcValue::TypeArray)
-//    {
-//      for (int32_t i = 0; i < topicList.size(); ++i)
-//      {
-//        ROS_ASSERT(topicList[i].getType() == XmlRpc::XmlRpcValue::TypeString);
-//        config_.input_scan_topics.push_back(static_cast<std::string>(topicList[i]));
-//
-//        ROS_DEBUG_STREAM("Parsed the scan topic: " << config_.input_scan_topics.back());
-//      }
-//
-//      config_.number_input_scans = config_.input_scan_topics.size();
-//    }
-//  }
-
   std::vector<std::string> topicList;
 
   if (pnh_.getParam("input_scans", topicList))
@@ -176,7 +156,11 @@ void ScanUnifierNode::messageFilterCallback(const sensor_msgs::LaserScan::ConstP
   current_scans.push_back(scan1);
   current_scans.push_back(scan2);
 
-  sensor_msgs::LaserScan unified_scan = unifyLaserScans(current_scans);
+  sensor_msgs::LaserScan unified_scan = sensor_msgs::LaserScan();
+  if (!unifyLaserScans(current_scans, unified_scan))
+  {
+    return;
+  }
 
   ROS_DEBUG("Publishing unified scan.");
   topicPub_LaserUnified_.publish(unified_scan);
@@ -189,7 +173,11 @@ void ScanUnifierNode::messageFilterCallback(const sensor_msgs::LaserScan::ConstP
   current_scans.push_back(scan2);
   current_scans.push_back(scan3);
 
-  sensor_msgs::LaserScan unified_scan = unifyLaserScans(current_scans);
+  sensor_msgs::LaserScan unified_scan = sensor_msgs::LaserScan();
+  if (!unifyLaserScans(current_scans, unified_scan))
+  {
+    return;
+  }
 
   ROS_DEBUG("Publishing unified scan.");
   topicPub_LaserUnified_.publish(unified_scan);
@@ -203,9 +191,9 @@ void ScanUnifierNode::messageFilterCallback(const sensor_msgs::LaserScan::ConstP
  * output:
  * @param: a laser scan message containing unified information from all scanners
  */
-sensor_msgs::LaserScan ScanUnifierNode::unifyLaserScans(std::vector<sensor_msgs::LaserScan::ConstPtr> current_scans)
+bool ScanUnifierNode::unifyLaserScans(std::vector<sensor_msgs::LaserScan::ConstPtr> current_scans, sensor_msgs::LaserScan &unified_scan)
 {
-  sensor_msgs::LaserScan unified_scan = sensor_msgs::LaserScan();
+//  sensor_msgs::LaserScan unified_scan = sensor_msgs::LaserScan();
   std::vector<sensor_msgs::PointCloud> vec_cloud;
   vec_cloud.assign(config_.number_input_scans, sensor_msgs::PointCloud());
 
@@ -219,13 +207,10 @@ sensor_msgs::LaserScan ScanUnifierNode::unifyLaserScans(std::vector<sensor_msgs:
       try
       {
         if (!listener_.waitForTransform(frame_, current_scans.at(i)->header.frame_id,
-                                        current_scans.at(i)->header.stamp, ros::Duration(3.0)))
+                                        current_scans.at(i)->header.stamp, ros::Duration(1.0)))
         {
-          continue;
-        }
-        else
-        {
-          ROS_WARN_STREAM("Scan unifier skipped scan with " << current_scans.at(i)->header.stamp << " stamp.");
+          ROS_WARN_STREAM("Scan unifier skipped scan with " << current_scans.at(i)->header.stamp << " stamp, because of missing tf transform.");
+          return false;
         }
 
         ROS_DEBUG("now project to point_cloud");
@@ -285,7 +270,7 @@ sensor_msgs::LaserScan ScanUnifierNode::unifyLaserScans(std::vector<sensor_msgs:
     }
   }
 
-  return unified_scan;
+  return true;
 }
 
 int main(int argc, char** argv)
--- a/cob_scan_unifier/src/scan_unifier_node.cpp
+++ b/cob_scan_unifier/src/scan_unifier_node.cpp
@@ -193,7 +193,6 @@ void ScanUnifierNode::messageFilterCallback(const sensor_msgs::LaserScan::ConstP
  */
 bool ScanUnifierNode::unifyLaserScans(std::vector<sensor_msgs::LaserScan::ConstPtr> current_scans, sensor_msgs::LaserScan &unified_scan)
 {
-//  sensor_msgs::LaserScan unified_scan = sensor_msgs::LaserScan();
   std::vector<sensor_msgs::PointCloud> vec_cloud;
   vec_cloud.assign(config_.number_input_scans, sensor_msgs::PointCloud());
 
--- a/cob_scan_unifier/src/scan_unifier_node.cpp
+++ b/cob_scan_unifier/src/scan_unifier_node.cpp
@@ -88,6 +88,7 @@ ScanUnifierNode::ScanUnifierNode()
     {
       typedef message_filters::sync_policies::ApproximateTime<sensor_msgs::LaserScan, sensor_msgs::LaserScan> SyncPolicy;
       synchronizer2_ = new message_filters::Synchronizer<SyncPolicy>(SyncPolicy(2), *message_filter_subscribers_.at(0), *message_filter_subscribers_.at(1));
+      // Set the InterMessageLowerBound to double the period of the laser scans publishing ( 1/{(1/2)*f_laserscans} ).
       synchronizer2_->setInterMessageLowerBound(0, ros::Duration(0.167));
       synchronizer2_->setInterMessageLowerBound(1, ros::Duration(0.167));
       synchronizer2_->registerCallback(boost::bind(&ScanUnifierNode::messageFilterCallback, this, _1, _2));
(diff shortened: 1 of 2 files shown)


cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -58,30 +58,79 @@
...
+      synchronizer3_ = new message_filters::Synchronizer<SyncPolicy>(SyncPolicy(3), *message_filter_subscribers_.at(0), *message_filter_subscribers_.at(1), *message_filter_subscribers_.at(2));
+      synchronizer3_->setInterMessageLowerBound(0, ros::Duration(0.167));
+      synchronizer3_->setInterMessageLowerBound(1, ros::Duration(0.167));
+      synchronizer3_->setInterMessageLowerBound(2, ros::Duration(0.167));
+      synchronizer3_->registerCallback(boost::bind(&ScanUnifierNode::messageFilterCallback, this, _1, _2, _3));
+      break;
+    }
+    default:
+      ROS_ERROR_STREAM(config_.number_input_scans << " topics have been set as input, but scan_unifier doesn't support this.");
+      return;
```
mgruhler, line 109: Hmm, I don't like that this can only be done for 2 or 3 scans.
This was better in the old implementation. Is there a way to achieve the same here?

mgruhler (changes requested):

cob_scan_unifier/src/scan_unifier_node.cpp
ghost: I don't like this too, but message_filters::TimeSynchronizer is implemented for a max of 9 scans, with a separate impl for every number-of-scanners. I tried but I didn't find a way to make it number-of-scanners-independent. I talked to ipa-fxm about this and he didn't know a better solution either.

mgruhler: To me, the documentation on the [wiki](http://wiki.ros.org/message_filters#Policy-Based_Synchronizer_.5BROS_1.1.2B-.5D) seems to state that we can sync up to 9 topics:

> C++: Up to 9 separate filters, each of which is of the form void callback(const boost::shared_ptr<M const>&). The number of filters supported is determined by the number of template arguments the class was created with.

I'd say, this should be fine then...

mgruhler: @ipa-fxm fyi
@mig-em please note if this is superseeding another PR and cross-referene.

mgruhler: Well, shouldn't we make it in a way then to be able to handle the nine scanners then? Use a vector of synchronizers, maybe?

mgruhler: alright, ignore this one: discussed in person.

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -123,150 +166,99 @@ void scan_unifier_node::getParams()
...
-            vec_laser_struct_.at(i).current_scan_msg.header.stamp, ros::Duration(3.0));
+        if (!listener_.waitForTransform(frame_, current_scans.at(i)->header.frame_id,
+                                        current_scans.at(i)->header.stamp, ros::Duration(3.0)))
+        {
+          continue;
+        }
+        else
+        {
+          ROS_WARN_STREAM("Scan unifier skipped scan with " << current_scans.at(i)->header.stamp << " stamp.");
+        }
```
mgruhler: Is this correct? We `continue` when `waitForTransform` times out and throw a warning if not. Or did I mix that up?

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -58,30 +58,79 @@
...
 
-  pthread_mutex_init(&m_mutex, NULL);
+  switch (config_.number_input_scans)
+  {
+    case 2:
+    {
+      typedef message_filters::sync_policies::ApproximateTime<sensor_msgs::LaserScan, sensor_msgs::LaserScan> SyncPolicy;
+      synchronizer2_ = new message_filters::Synchronizer<SyncPolicy>(SyncPolicy(2), *message_filter_subscribers_.at(0), *message_filter_subscribers_.at(1));
+      synchronizer2_->setInterMessageLowerBound(0, ros::Duration(0.167));
+      synchronizer2_->setInterMessageLowerBound(1, ros::Duration(0.167));
```
mgruhler: How did you choose this number?

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -58,30 +58,79 @@
...
+      synchronizer3_->registerCallback(boost::bind(&ScanUnifierNode::messageFilterCallback, this, _1, _2, _3));
+      break;
+    }
+    default:
+      ROS_ERROR_STREAM(config_.number_input_scans << " topics have been set as input, but scan_unifier doesn't support this.");
+      return;
+  }
 
-  initLaserScanStructs();
+//  synchronizer_->registerCallback(&ScanUnifierNode::messageFilterCallback);
```
mgruhler: remove, this is not used.

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -123,150 +166,99 @@ void scan_unifier_node::getParams()
...
       }
     }
     ROS_DEBUG("Creating message header");
-    unified_scan.header = vec_laser_struct_.at(0).current_scan_msg.header;
-    unified_scan.header.frame_id = "base_link";
+    unified_scan.header = current_scans.at(0)->header;
+    unified_scan.header.frame_id = frame_;
     unified_scan.angle_increment = M_PI/180.0/2.0;
     unified_scan.angle_min = -M_PI + unified_scan.angle_increment*0.01;
     unified_scan.angle_max =  M_PI - unified_scan.angle_increment*0.01;
```
mgruhler, line 227: This might actually be the reason for #249 

Also, the above restricts this unifier to be used with a spacing fo 0.5°. This should probably be adapted as well.
Something like "biggest common denominator" but in float ;-)
Added it to #249

mgruhler (changes requested):

cob_scan_unifier/src/scan_unifier_node.cpp
fmessmer: =0.5*(1/12) // half of expected scanner rate at 12Hz (I guess)

cob_scan_unifier/src/scan_unifier_node.cpp
fmessmer: For the sake of completeness: [Synchronizer-API](http://docs.ros.org/indigo/api/message_filters/html/c++/classmessage__filters_1_1TimeSynchronizer.html) is implemented in the same (ugly) way...

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -58,30 +58,79 @@
...
+    {
+      delete(synchronizer2_);
+      break;
+    }
+    case 3:
+    {
+      delete(synchronizer3_);
+      break;
+    }
+  }
```
fmessmer: I guess there is actually no need to switch...you could just `delete` both...

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -90,17 +139,11 @@ scan_unifier_node::~scan_unifier_node(){}
...
-
-  if(!pnh_.hasParam("loop_rate"))
-  {
-    ROS_WARN("No parameter loop_rate on parameter server. Using default value [100.0].");
-  }
-  pnh_.param("loop_rate", config_.loop_rate, (double)100.0);
-
   XmlRpc::XmlRpcValue topicList;
 
+  // TODO short parsing
```
fmessmer: Link to list retrieval: http://wiki.ros.org/roscpp/Overview/Parameter%20Server#Retrieving_Lists

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -123,150 +166,99 @@ void scan_unifier_node::getParams()
...
-{
-  return config_.loop_rate;
+  if(!pnh_.hasParam("frame"))
+  {
+    ROS_WARN("No parameter frame on parameter server. Using default value [base_link].");
+    frame_ = "base_link";
+  }
+  else
+  {
+    pnh_.getParam("frame", frame_);
```
fmessmer: just use "default version" (`nh.param<std::string>("default_param", default_param, "default_value");`) ?

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -123,150 +166,99 @@ void scan_unifier_node::getParams()
...
-/**
- * @function initLaserScanStructs
- * @brief initialize a vector of laser scan structs (member variable vec_laser_struct_) with a given number
- * (from parameter server, stored in config_ struct)
- *
- * input: -
- * output: -
- */
-void scan_unifier_node::initLaserScanStructs()
+void ScanUnifierNode::messageFilterCallback(const sensor_msgs::LaserScan::ConstPtr& first_scanner, const sensor_msgs::LaserScan::ConstPtr& second_scanner)
```
fmessmer: I'd prefer `scan1` instead of `first_scanner`, `scan2` instead of `second_scanner`....

cob_scan_unifier/src/scan_unifier_node.cpp
fmessmer: I am confused as well....
With the Synchronizer we never should need to skip a scan any more...that's the whole point....

cob_scan_unifier/src/scan_unifier_node.cpp
mgruhler: Hmm, is it safe to assume this is the lower bound? Or would we want to merge scans with a frequency <12Hz as well?

fmessmer: Is 12Hz a configurable parameter for the scanners?
Then we at least should make the computation of the bounds based on a parameter as we might unify scans from other sensors with different publish rate as well...

mgruhler: To be honest, I'm not sure. There is a [parameter about the publish rate in the S300 driver](https://github.com/ipa320/cob_driver/blob/indigo_dev/cob_sick_s300/ros/src/cob_sick_s300.cpp#L145-L146). But: afaik, this is [not respected](https://github.com/ipa320/cob_driver/blob/indigo_dev/cob_sick_s300/ros/src/cob_sick_s300.cpp#L259-L261). The new S300 Expert do have an output of about 25Hz (phyiscal max of the scanner) with our driver.

The frequency of the data output depends on the configuration in the scanner itself.

Also, other scanners have other frequencies... And we could actually try to merge scanners with different frequencies. But this gets ugly.

fmessmer: let's make it work for our use-case....either 2 s300 (raw) or 3 s300 (cob4)...then test it...eventually it's **cob**_scan_unifier anyway
(but add a comment about where the formula comes from, i.e. relate it to expected s300 publish rate)

mgruhler: fine for me. Then we can stick to 12Hz (even though the scanner itself could be configured to send out with less frequency without respecting anything in the driver. we just should not do that).

Then we can stick with the 0.5° spacing in my comment below as well...

fmessmer: Well, I meant **after** successful testing we could at least introduce parameters to make the scan_unifier node configurable...but those parameters would be scan_unifier parameters and not related to parameters (of the same meaning) of the scanner driver, though

mgruhler: Yes, I fully understand this. I just wanted to point out that this is in the end not independent (high expected frequency in the unifier + low frequency in the scanner = no unified scan).

But I'm sure I'm preaching to the choir here :-)

cob_scan_unifier/src/scan_unifier_node.cpp
ghost: Synchronizer does only sync the messages, so it doesn't make any difference for the tf problem...

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -90,183 +126,130 @@ scan_unifier_node::~scan_unifier_node(){}
...
+//      {
+//        ROS_ASSERT(topicList[i].getType() == XmlRpc::XmlRpcValue::TypeString);
+//        config_.input_scan_topics.push_back(static_cast<std::string>(topicList[i]));
+//
+//        ROS_DEBUG_STREAM("Parsed the scan topic: " << config_.input_scan_topics.back());
+//      }
+//
+//      config_.number_input_scans = config_.input_scan_topics.size();
+//    }
+//  }
```
mgruhler: This could/should be removed...

mgruhler (changes requested):

cob_scan_unifier/src/scan_unifier_node.cpp
mgruhler: Okay, this is fine.

But: Isn't this still the wrong way around?

fmessmer: Well, now that I'm looking at it:
**We** are handing over the `tf::TransformListener` object to the `projector_`. So in case all the `waitForTransform` magic we do in order to make the `transformLaserScanToPointCloud` work....maybe we just need to make our `listener_` object work!

Because currently it's using the default constructor (http://docs.ros.org/indigo/api/tf/html/c++/classtf_1_1TransformListener.html), i.e. default values for `max_cache_time`...maybe we can tweak something here...what also helps qutie often is "wait a little" at the end of the constructor to let the listener_ fill its cache...as we are seeing this problem only in the very beginning...

Also, is this `ROS_WARN_STREAM` still seen in the log? Because maybe the better time synchronization now does not give us messy timestamps as the "old" synchronization might have...

Last not least, in case the `!waitForTranform` case cannot be avoided by any means...should we return, i.e. skip all scans of current cycle or only skip the scan that failed, i.e. incomplete scan unified

fmessmer: > Is this correct? We `continue` when `waitForTransform` times out and throw a warning if not. Or did I mix that up?

ja is wrong!

mgruhler (approved):

PR #326, opened 2017-03-08
setting current for realistic power_state calculations 

souravran:
made changes which only sets the current that in turn is used by power_aggregator for relative_remaining_capacity calculation


floweisshardt (approved):

fmessmer:
fix https://github.com/ipa320/msh/issues/636

PR #328, opened 2017-03-10
fix cpu usage of the mimic node

NadiaHG:
needed for the new vlc version

fix https://github.com/ipa320/cob4/issues/244


NadiaHG:
test on cob4-5: OK

fmessmer (approved):
So the `--video-wallpaper` is not supported in the latest vlc version? Or is this argument just causing high load?

floweisshardt (approved):

PR #329, opened 2017-03-15
Fix/diagnostic rate

### LINKED ISSUE #207: base_drive_chain's diagnostics message is published with 100 Hz
The base_drive_chain diagnostics message is published with 100Hz. This is not necessary and should be reduced. I think every half second or second is absolutely enough and could be easily done with a simple ros timer.
@ipa-mig What do you think?

ghost:
fix ipa320/cob_driver#207


fmessmer:
@ipa-mig 
why do we touch this package?
where is it still used? raw? I'd prefer to continue/finalize @ipa-jba's work on using ros_canopen on raw....see https://github.com/ipa320/cob_robots/issues/422 and https://github.com/ipa320/cob_robots/pull/609 for the raw3-5 example

mgruhler (approved):
please do a squash merge

PR #330, opened 2017-03-28
renamed max_voltage to more descriptive voltage_divider_factor

jabbenseth:
depends on https://github.com/ipa320/cob_robots/pull/623
sorry about changes in whitespace

### DIFF
--- a/cob_phidget_power_state/scripts/power_state_phidget.py
+++ b/cob_phidget_power_state/scripts/power_state_phidget.py
@@ -17,8 +17,10 @@ class PowerStatePhidget():
         self.current = None
         self.last_update = rospy.Time(0)
         self.charging = False
-
-        self.voltage_divider_factor = rospy.get_param("~voltage_divider_factor", 59.5)
+        try:
+            self.voltage_divider_factor = rospy.get_param("~voltage_divider_factor")
+        except KeyError:
+            raise KeyError("Parameter \"~voltage_divider_factor\" not found on parameter server.")
         self.voltage_full = rospy.get_param("~voltage_full", 52.0)
         self.voltage_empty = rospy.get_param("~voltage_empty", 38.0)
         self.current_max = rospy.get_param("~current_max", 30.0)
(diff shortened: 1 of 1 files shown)


cob_phidget_power_state/scripts/power_state_phidget.py
```diff
@@ -1,23 +1,24 @@
...
+    VOLTAGE_COLLECTION_TIME = 6.0  # sec
 
     def __init__(self):
         self.voltage = None
         self.current = None
         self.last_update = rospy.Time(0)
         self.charging = False
 
-        self.voltage_max = rospy.get_param("~voltage_max", 59.5)
+        self.voltage_divider_factor = rospy.get_param("~voltage_divider_factor", 59.5)
```
floweisshardt: if this is no more a maximum voltage is there a better default value? If not, I would make this a non-optional parameter and thow an error if missing.

floweisshardt (changes requested):

floweisshardt (approved):

fmessmer:
imho, the descriptiveness of `voltage_divider_factor` isn't much better...division and factor sounds kind of a contradiction in itself...and it's not used within a division...how  about `voltage_scaling_factor....
just my five cents.....I leave this to you guys...

jabbenseth:
The division part comes from the hardware voltage divider. The resulting values are multiplied. But I see your point. What about `voltage_divider_ratio` and changing the values and computation to `1/voltage_divider_factor` to keep it consistent with the general case described by wikipedia: https://en.wikipedia.org/wiki/Voltage_divider
?

fmessmer:
No need to make it more complicated than it is....if everybody is ok, go for it...(add a descriptive comment, if you like)...

jabbenseth:
@ipa-bnm you implemented the node. whats your opinion?

fmessmer (approved):

benmaidel:
@ipa-jba too late ;)

PR #331, opened 2017-03-28
[cob_sick_s300] Added private node handle for parameters.

tonybaltovski:
Added a private node handle for parameters to allow for multiple lasers to be used.

### DIFF
--- a/cob_sick_s300/ros/src/cob_sick_s300.cpp
+++ b/cob_sick_s300/ros/src/cob_sick_s300.cpp
@@ -85,7 +85,7 @@ class NodeClass
 	//
 	public:
 
-		ros::NodeHandle nh;
+		ros::NodeHandle nh, pnh;
 		// topics to publish
 		ros::Publisher topicPub_LaserScan;
 		ros::Publisher topicPub_InStandby;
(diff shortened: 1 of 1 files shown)


fmessmer:
While I see the use-case, merging this would require us to reconsider our current configuration in https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_hardware_config/cob4-2/config/base_laser_left.yaml and the launch file https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/drivers/sick_s300.launch because currently we also have the laser filter params in the same yaml....

tonybaltovski:
I will follow suite and use the group namespace.  I thought it was a overlooked as the private node handle was commented out [here](https://github.com/ipa320/cob_driver/blob/indigo_dev/cob_sick_s300/ros/src/cob_sick_s300.cpp#L121).

fmessmer:
So, I'll close this PR unmerged....

mgruhler:
I know, I'm quite late to the party. But thanks @tonybaltovski this has bothered me for quite a while as well.

I'm not quite sure if, or if yes, when, we'll tackle this. This is on my list as well, but doe to the configuration issues I haven't fixed it yet...

tonybaltovski:
@ipa-mig No worries.  If you do change it, let me know.  Maybe, it could be considered for kinetic?

PR #334, opened 2017-05-17
[cob_sick_s300] adjust range_max

mgruhler:
@ipa-srd FYI

### DIFF
--- a/cob_sick_s300/ros/src/cob_sick_s300.cpp
+++ b/cob_sick_s300/ros/src/cob_sick_s300.cpp
@@ -291,7 +291,7 @@ class NodeClass
 			laserScan.header.frame_id = frame_id;
 			laserScan.angle_increment = vdAngRAD[start_scan + 1] - vdAngRAD[start_scan];
 			laserScan.range_min = 0.001;
-			laserScan.range_max = 30.0;
+			laserScan.range_max = 29.5; // though the specs state otherwise, the max range reported by the scanner is 29.96m
 			laserScan.time_increment = (scan_duration) / (vdDistM.size());
 
 			// rescale scan


stefandoerr:
Could lead to some "surprising" behavior though. For instance, we will not see the max range measurements in RVIZ anymore. Also, I'm not sure if some nodes "rely" on this bug. For instance, the `costmap` discards max range measurements. Thereby, there will be no ray tracing and free space clearing with those measurements. However, I support making this change since it is imo the correct way.

cob_sick_s300/ros/src/cob_sick_s300.cpp
```diff
@@ -291,7 +291,7 @@ class NodeClass
 			laserScan.header.frame_id = frame_id;
 			laserScan.angle_increment = vdAngRAD[start_scan + 1] - vdAngRAD[start_scan];
 			laserScan.range_min = 0.001;
-			laserScan.range_max = 30.0;
+			laserScan.range_max = 29.5; // though the specs state otherwise, the max range reported by the scanner is 29.96m
```
stefandoerr, line 294: why not set it to `29.96` then?

mgruhler: To have a defined boundary. The max output of the scanner (I observed yesterday during some tests) was 2996mm, but due to the float conversion I figured it is better to have some clearance.

In the end, for localization/mapping tasks, we will probably limit the range anyways to something like 15m (at least indoors...)

mgruhler:
I agree with @ipa-srd about the behavior. We should probably check/investigate if this has some influence on the navigation, especially with the CoB4 settings/environments @ipa-fmw @ipa-fxm

fmessmer:
> though the specs state otherwise

so what is in the specs?

what's the story behind this? what happens with `range_max = 30.0`, that doesn't with `range_max = 29.whatever`?

 > In the end, for localization/mapping tasks, we will probably limit the range anyways to something like 15m (at least indoors...)

so does it make sense to make this a parameter?

 > We should probably check/investigate if this has some influence on the navigation, especially with the CoB4 settings/environments

definitely test in market temporarily, please

stefandoerr:
The problem is that the msg states a max range value of 30 m. But an actual max range measurement appears in the range vec with an value something 29.96 m. So in the end, those range measurements are not classified as max range measurements since they are below the threshold. In SLAM for instance, this leads to mapping those points as objects although they are no real reflections

mgruhler:
@ipa-srd pretty much covered it.

`range_max` here should not be a parameter. The limitation happens in the respective localization/mapping/higher level nodes.

mgruhler:
@ipa-fmw do you want to coordinate the testing?

fmessmer:
well, maybe it's device-specific...thus it could be a candidate for the laser.yaml...please check with several scanners we have around...

please also look into the message comming from the gazebo plugin in simulation...to guarantee identical behavior hw vs. sim

mgruhler:
@ipa-fxm I don't know if I'll be able to check this tomorrow. I need a scanner and 30+m of free fov.
I tested with a S300 Pro (old version) not mounted on a robot. I have no other unmounted scanner with cables... Specs say, range is 30m.

Do you want me to adjust [the gazebo plugin](https://github.com/ipa320/cob_common/blob/indigo_dev/cob_description/urdf/sensors/sick_s300_laser.gazebo.xacro#L22) config accordingly? I don't know how this handles max range measurements. But if there is something wrong with sim, the problem is already there... (BTW, I just noticed the resolution in sim is set to 0.03m, however, in reality, it is 0.01, One could also increase the noise, which is typically a bit higher)

For reference, [another implementation](https://github.com/bohlender/sicks300/blob/master/src/sicks300.cpp#L86) sets the `range_max` to 29m...

mgruhler:
also, this is not (should not) be device specific... in laser scanners, `>=range_max` means "invalid", and should be discarded. Though if there is a scanner actually delivering a range of `30.03m` the new `range_max` does actually not change anything...

mgruhler:
"Shooting" out the window
Old behaviour:
old
New behaviour:
new
Simulation old (30m) in empty world:
sim_old
Simulation new (10m) in empty world:
sim_new


IMO, simulation is wrong. You cannot specify the maximum range and the `range_max` seperately. This leads to almost all measurements being in the valid range. You see some "gaps" in the simulated scan, i.e. where it is exactly the max range (10/30m). This gets filtered, the points shown are at (10/30m-epsilon) (e.g. 9.996..., 29.996...), probably due to rounding, and are thus shown.

I don't mind if you merge this or not. But this would be the correct way.
Some of our students actually started to implement special handling for the S300 because of this...

fmessmer:
Ok, so in simulation we'd keep 30.0 as it is not just used for filling the `range_max` field in the message, but also as the range for which ray-tracing is performed...

fmessmer (approved):
given the discussion and explanations, I approve...

@ipa-fmw or @ipa-bnm please merge in case you are happy with it

mgruhler:
But please remember to test as well in reality! Works in simulation and in small environments. but what about larger ones...

fmessmer:
Have any more tests been performed with this feature branch? Did anything come up?
@ipa-fmw @ipa-mig feel free to merge if nothing came up...

mgruhler:
AFAIR not. IMO, this should be tested by @ipa-fmw on paul. This is currently the one big environment where we will actually run into problems...

fmessmer:
@ipa-fmw ping?

floweisshardt:
didn't have time to test on paul yet... can somebody take over?

ipa-owner-jba:
close. move to #358

PR #337, opened 2017-05-31
fix for the flickering in playback, caused when an emotion is set.

souravran:


souravran:
PR should address https://github.com/ipa320/msh/issues/714#issuecomment-302705555

cob_mimic/src/mimic_node.py
```diff
@@ -60,6 +60,7 @@
 import sys
 import os
 import subprocess
+import vlc
```
fmessmer, line 63: is this python module available via `pip`?
adding the whole modules seems overkill...

EDIT:
seems available via `sudo pip install python-vlc`...maybe we can add a `rosdep` key to rosdistro...

fmessmer (changes requested):

cob_mimic/src/mimic_node.py
souravran: Actually, we tried doing that but I think some dependencies were not met.
@ipa-nhg please correct me, if I am wrong.

souravran: Just had a chat with @ipa-nhg we could go the ```rosdep``` way and install vlc python bindings.

cob_mimic/src/mimic_node.py
```diff
@@ -69,69 +70,100 @@
...
+        if self.repeat and self.emotion:
+            self.player.set_rate(float(self.playback_speed))
+            file_location = '/tmp/mimic/' + self.emotion + '.mp4'
+            while self.repeat:
+                self.media = self.vlc_instance.media_new(file_location)
+                self.media.get_mrl()
+                self.player.set_media(self.media)
+                self.player.play()
+                self.repeat -= 1
+                rospy.sleep(4)
```
floweisshardt: instead of a hardcoded sleep I'd like to have a blocking version here. should be possible to get a `playing` flag from VLC.

```
play()
while playing:
    sleep(0.1)
```


ALTERNATIVE: get the length of the video and sleep for the length of the video mulitplied by `speed`

cob_mimic/src/mimic_node.py
```diff
@@ -69,69 +70,100 @@
...
+                rospy.logerror("File not found: %s", self.default_mimic_file)
+                return
+            # run single instance of the media player
+            self.vlc_instance = vlc.Instance('--ignore-config', '--mouse-hide-timeout=0', '-q', '--no-osd', '-L', '--one-instance', '--playlist-enqueue', '--no-video-title-show')
+            self.player = self.vlc_instance.media_player_new()
+            self.player.set_fullscreen(int(True))
+        except Exception as e:
+            rospy.logerror("Something went wrong while setting up media playback: %s", str(e))
+
+    def play_defualt_emotion(self):
```
floweisshardt: typo

floweisshardt (changes requested):

cob_mimic/src/mimic_node.py
```diff
@@ -69,69 +70,100 @@
...
+        self._ss = rospy.Service('~set_mimic', SetMimic, self.service_cb)
+        self._as = actionlib.SimpleActionServer('~set_mimic', cob_mimic.msg.SetMimicAction, execute_cb=self.action_cb, auto_start = False)
+        self._as.start()
+        self.play_defualt_emotion()
+        rospy.spin()
+
+    def setup_playback(self):
+        try:
+            if not os.path.isfile(self.default_mimic_file):
+                rospy.logerror("File not found: %s", self.default_mimic_file)
```
NadiaHG: typo `rospy.logerr`

cob_mimic/src/mimic_node.py
```diff
@@ -69,69 +70,100 @@
...
+                return
+            # run single instance of the media player
+            self.vlc_instance = vlc.Instance('--ignore-config', '--mouse-hide-timeout=0', '-q', '--no-osd', '-L', '--one-instance', '--playlist-enqueue', '--no-video-title-show')
+            self.player = self.vlc_instance.media_player_new()
+            self.player.set_fullscreen(int(True))
+        except Exception as e:
+            rospy.logerror("Something went wrong while setting up media playback: %s", str(e))
```
NadiaHG: typo `rospy.logerr`

cob_mimic/src/mimic_node.py
```diff
@@ -69,69 +70,100 @@
...
+            self._as.set_succeeded()
+        else:
+            self._as.set_aborted()
+
+    def set_mimic(self, mimic, speed, repeat):
+        rospy.loginfo("Mimic: %s", mimic)
+        file_location = '/tmp/mimic/' + mimic + '.mp4'
+        print mimic
+        if(not os.path.isfile(file_location)):
+            rospy.logerror("File not found ")
```
NadiaHG: typo `rospy.logerr`

cob_mimic/src/mimic_node.py
NadiaHG: this works:

```
        rospy.sleep(0.1)
        while self.player.is_playing() == 1:
                rospy.sleep(0.1)
```

You need a small sleep after play an before check if it is playing, else is_playing will return with false. This loop could be added after each self.player.play() line 144 and 151

See the commit https://github.com/ipa-nhg/cob_driver/commit/7bdfa83c65476fb8b98b1a783ec64b410a5d1d04

NadiaHG (changes requested):

PR #344, opened 2017-07-21
[mimic] ported mimic from python to c++

### LINKED ISSUE #340: [cob_mimic] consider c++ implementation
motivated by discussion https://github.com/ipa320/care-o-bot/issues/42#issuecomment-316046500

The ROS release system does not support `python-pip` dependencies...thus, we cannot release `cob_mimic` with a dependency on `python-vlc-pip`.
We should implement `cob_mimic` in C++ using [`libvlc-dev`](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L2987-L2991) which is provided by `apt` which is supported by the ROS release system.

benmaidel:
- fixes #340


benmaidel:
Tested on cob4-7

floweisshardt (approved):
tested locally: OK

fmessmer (approved):
tested in sim :+1:

PR #345, opened 2017-07-25
guarantee unique copy destinations

fmessmer:
problem was "permission denied" when accessing "/tmp/mimic" from different users...without having rebooted (i.e. cleaned /tmp)


floweisshardt (approved):
tested with multiple users on paul-stuttgart: OK

PR #347, opened 2017-07-31
remove cob_relayboard

floweisshardt:
- fixes parts of https://github.com/ipa320/care-o-bot/issues/46
- needed for https://github.com/ipa320/cob_robots/pull/696


fmessmer:
do not merge (yet)
raw3-1 still use the fake node as it does not have phidgets https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/robots/raw3-1.xml#L33-L37

the `relayboard_sim` node could be moved to https://github.com/ipa320/cob_command_tools/tree/indigo_dev/cob_helper_tools/scripts maybe?

fmessmer:
a proper `flexisoft_sim` node has been added in ipa320/cob_substitute#46
thus, cob4's do not require cob_relayboard anymore...

still raw3-1 issue persists

floweisshardt:
I wouldn't add it to cob_helper_tools as it has nothing to do woth the rest in this package and would only add additional dependencies.

In my opinion raw3-1 does not need the simulated node as soon as we're using canopen with the cob_omni_driver_controller instead of the outdated cob_undercarriage_control. Until this is done, I'm fine with using the sim node from sick_flexisoft for raw3-1 as we have this dependency anyway in cob_bringup.

fmessmer:
**Do not merge** (yet)
will be investigated in ipa320/cob_robots#703

fmessmer:
How do we deal with this?
See https://github.com/ipa320/care-o-bot/issues/46#issuecomment-327392494

fmessmer:
let's keep these packages for the time being...we can still remove them once they require more maintenance effort in the future

PR #351, opened 2017-08-02
add tinyxml to package.xml

mikaelarguedas:
while investigating packages depending on `tinyxml` I noticed that this one didn't declare the dependency in its `package.xml` while finding it in its cmake code.


ipa-rmb (approved):
Thanks for finding that. It did not harm since these files are only relevant for the Windows version, but this way it is correct.

fmessmer (approved):

PR #361, opened 2017-11-28
set relative remaining capacity

fmessmer:
I found setting the `relative_remaining_capacity` via service more intuitive than setting the `remaining_capacity` when testing low-battery behavior in simulation for MSH
This is due to all battery-monitor conditions take `relative_remaining_capacity` into account, but `relative_remaining_capacity` is computed only in the [`power_state_aggregator`](https://github.com/ipa320/cob_driver/blob/indigo_dev/cob_bms_driver/src/power_state_aggregator.py#L81)

### DIFF
--- a/cob_bms_driver/src/fake_bms.py
+++ b/cob_bms_driver/src/fake_bms.py
@@ -24,7 +24,7 @@ from diagnostic_msgs.msg import DiagnosticArray, DiagnosticStatus
 class FakeBMS(object):
     def __init__(self):
         self.srv_current              = rospy.Service('~set_current', SetFloat, self.current_cb)
-        self.srv_remaining_capacity   = rospy.Service('~set_remaining_capacity', SetFloat, self.remaining_capacity_cb)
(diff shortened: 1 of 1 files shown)


ipa-rmb (dismissed):

fmessmer:
@ipa-rmb @ipa-jba @ipa-fez @ipa-bnm @ipa-fmw 
I'd need another review here

cob_bms_driver/src/fake_bms.py
```diff
@@ -74,9 +74,9 @@ def timer_cb(self, event):
 
     def timer_consume_power_cb(self, event):
         # emulate the battery usage based on the current values
-        self.remaining_capacity += (self.current/self.poll_frequency)/3600   
-        if self.remaining_capacity <= 0:
-            self.remaining_capacity = 0
+        self.remaining_capacity += round(((self.current/self.poll_frequency)/3600), 5)
```
ipa-fez: Won't this prevent the capacity from updating if the delta is very small and always rounded to 0? Is there a reason you need to round?

ipa-fez (changes requested):

fmessmer:
thx for the comments...please approve again :wink:

floweisshardt (approved):

PR #362, opened 2017-12-06
[cob_sound] catch if filename is empty or invalid

floweisshardt:
without that fix sound node will crash if 
- [x] filename is empty or 
- [ ] file does not exist or
- [ ] file is not playable

### DIFF
--- a/cob_sound/ros/src/sound.cpp
+++ b/cob_sound/ros/src/sound.cpp
@@ -266,15 +266,22 @@ public:
     return true;
   }
 
-  bool play(std::string filename, std::string message)
+  bool play(std::string filename, std::string &message)
   {
     bool ret = false;
     if (mute_)
     {
-      message = "Sound is set to mute. You will hear nothing."; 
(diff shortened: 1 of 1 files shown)


floweisshardt:
@HannesBachter can you check if we can handle "file does not exist" and "file is not playable"?

for that we should use libvlc to querry if file is playable.
we've already tried to use `libvlc_media_player_will_play()` https://www.videolan.org/developers/vlc/doc/doxygen/html/group__libvlc__media__player.html#ga3c38df27467a03f96f9ed0f6b585baaa but this does not play files at all (e.g. mp3 files)

another thread which might be worth to try is https://stackoverflow.com/questions/28263565/libvlc-check-if-media-location-is-valid which tries to register a callback which will fire on an error event. In that case we could call `fade_out` to stop the player (and probably have to take care about canceling the action or returnnin the service)

fmessmer (approved):
I created a new issue https://github.com/ipa320/cob_driver/issues/369 for the remaining cases...but current PR can get merged

Issue #369, opened 2018-04-18
[cob_sound] bullet-proofing bad file cases


fmessmer:
enhance cob_sound node to not crash in the following cases:
 - file does not exist
 - file is not playable

follow up of https://github.com/ipa320/cob_driver/pull/362

see hints in https://github.com/ipa320/cob_driver/pull/362#issuecomment-349585200

fmessmer:
> file is not playable

vlc does not care much about a file not being playable...I was able to call jpg, bmp, mp3, mp4, pdf, docx....either it played or did not...but it didn't crash...

PR #371, opened 2018-05-04
allow to play non-default mimics by specifying full filepath

fmessmer:
this is a minimal patch that allows to play mimics (or any other *.mp4 video file) from any file location
it does not change the default behavior at all

you only need to specify the full file path in the goal/request instead of just specifying default mimics like `happy`, `sleeping` or `confused`, i.e. `/home/fxm/Videos/my_video.mp4`

@HannesBachter @ipa-bnm @ipa-fmw 
we could use this for HDG's QR visualization and/or error_mimic (as long as we do not yet reach the pop-up solution)

### DIFF
--- a/cob_mimic/src/mimic_node.cpp
+++ b/cob_mimic/src/mimic_node.cpp
@@ -196,9 +196,17 @@ private:
         // check if mimic exists
         if ( !boost::filesystem::exists(filename) )
         {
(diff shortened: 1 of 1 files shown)


ipa-rmb (approved):

fmessmer:
works on robot as expected...
further bullet-proofing will be done in #369

Issue #338, opened 2017-06-02
[cob_sick_s300] frame rate low with RS-422


ehocquard:
Hello everyone,

I am working with sick s300 advanced sensors for a UAV project. My system is designed with ROS and thanks to your work I have compatibility with sick and ROS.

The sick is connected via a RS-422 interface set up at 500k baud on continuous data transfer, it works.

However, I do observe a rate of scan transfer (on the topic) of **12 Hz** only. [EDIT] with a `publish_frequency` set at **50**
I was wondering if anyone had similar results, it sounds a bit low rate  for a scanner that will be use for autonomous driving. Is this limited by the RS-422 or there is another bottleneck that I don't see?

If anyone can share experience on that I'll be grateful!

Best regards,
Etienne H.

PS: sorry if raising an issue for this is not appropriate!

fmessmer:
there is a parameter `publish_frequency` which defaults to 12 :wink: 
https://github.com/ipa320/cob_driver/blob/indigo_dev/cob_sick_s300/ros/src/cob_sick_s300.cpp#L146

ehocquard:
Indeed, thank you for pointing it out.

However changing the parameter does not change the rate of the received message, hence my question/issue on similar experience and maximum speed reached by others.

Log with a publish frequency set to 50:
"average rate: 12.415
	min: 0.028s max: 2.572s std dev: 0.06747s window: 1366
average rate: 12.417
	min: 0.028s max: 2.572s std dev: 0.06715s window: 1379
average rate: 12.420
	min: 0.028s max: 2.572s std dev: 0.06683s window: 1392
average rate: 12.422
	min: 0.028s max: 2.572s std dev: 0.06655s window: 1404
average rate: 12.424
	min: 0.028s max: 2.572s std dev: 0.06624s window: 1417
average rate: 12.427
	min: 0.028s max: 2.572s std dev: 0.06594s window: 1430
average rate: 12.429
	min: 0.028s max: 2.572s std dev: 0.06567s window: 1442
average rate: 12.431
	min: 0.028s max: 2.572s std dev: 0.06537s window: 1455
^Caverage rate: 12.433
	min: 0.028s max: 2.572s std dev: 0.06513s window: 1466"

mgruhler:
Hi @ehocquard 

There are several things to consider here:
[...]

PR #373, opened 2018-06-01
fixed limit for publish frequency

### LINKED ISSUE #372: s300 publish rate throttling does not work

kalectro:
- fixes #372 
 - new source code files added (proper [APACHE license header](https://github.com/ipa320/setup/tree/master/templates) **must** be used)

### DIFF
--- a/cob_sick_s300/ros/src/cob_sick_s300.cpp
+++ b/cob_sick_s300/ros/src/cob_sick_s300.cpp
@@ -220,10 +220,12 @@ class NodeClass
 		// other function declarations
 		void publishLaserScan(std::vector<double> vdDistM, std::vector<double> vdAngRAD, std::vector<double> vdIntensAU, unsigned int iSickTimeStamp, unsigned int iSickNow)
 		{
+			ros::Time now = ros::Time::now();
+			
 			// do not publish new scan if it would exceed the maximum publish frequency
-			if(ros::Time::now()-timestamp_last_published_ < ros::Duration(1./publish_frequency))
+			if(now-timestamp_last_published_ < ros::Duration(1./publish_frequency))
 				return;
-			timestamp_last_published_ = ros::Time::now();
+			timestamp_last_published_ = now;
 
 			// fill message
 			int start_scan, stop_scan;
@@ -234,7 +236,7 @@ class NodeClass
 			// Sync handling: find out exact scan time by using the syncTime-syncStamp pair:
 			// Timestamp: "This counter is internally incremented at each scan, i.e. every 40 ms (S300)"
 			if(iSickNow != 0) {
-				syncedROSTime = ros::Time::now() - ros::Duration(scan_cycle_time);
+				syncedROSTime = now - ros::Duration(scan_cycle_time);
 				syncedSICKStamp = iSickNow;
 				syncedTimeReady = true;
 
@@ -247,9 +249,9 @@ class NodeClass
 				double timeDiff = (int)(iSickTimeStamp - syncedSICKStamp) * scan_cycle_time;
 				laserScan.header.stamp = syncedROSTime + ros::Duration(timeDiff);
 
-				ROS_DEBUG("Time::now() - calculated sick time stamp = %f",(ros::Time::now() - laserScan.header.stamp).toSec());
+				ROS_DEBUG("Time::now() - calculated sick time stamp = %f",(now - laserScan.header.stamp).toSec());
 			} else {
-				laserScan.header.stamp = ros::Time::now();
+				laserScan.header.stamp = now;
 			}
 
 			// fill message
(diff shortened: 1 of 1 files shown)


ipa-rmb:
Many thanks for finding this bug. I was so free to append a change on the variable name loop_rate_, which was no loop rate, to make it more meaningful for the given purpose.

ipa-rmb (dismissed):

cob_sick_s300/ros/src/cob_sick_s300.cpp
```diff
@@ -220,9 +220,10 @@ class NodeClass
 		// other function declarations
 		void publishLaserScan(std::vector<double> vdDistM, std::vector<double> vdAngRAD, std::vector<double> vdIntensAU, unsigned int iSickTimeStamp, unsigned int iSickNow)
 		{
-			if(ros::Time::now()-loop_rate_.now()>=ros::Duration(1./publish_frequency))
```
fmessmer, line 223: I think we should set a variable
```
ros::Time now = ros::Time::now();
```
right at the beginning of the publish function  and then use `now` throughout the function instead of getting new times everytime

fmessmer (changes requested):

ipa-rmb (approved):

fmessmer:
@kalectro do you still see yourself in this PR? :wink:

ipa-rmb:
@ipa-fxm are you fine with the PR as well? Please approve then. I will merge after @kalectro and you confirmed.

ipa-rmb:
@fmessmer if you do not have any further objections, I will merge it tomorrow.

fmessmer:
@ipa-rmb @kalectro 
I have a bunch of PRs to test on HW next week...
thus, let me test this one, too...I'll merge after testing

fmessmer (changes requested):
I tested this PR today...it seems there is a need for a little more discussion:

on our robot we have `publish_frequency` set to `12Hz` and - with current `ipa320/indigo_dev` - we receive a topic rate of `25Hz` (using `rostopic hz /base_laser_front/scan_raw`)
with this PR by @kalectro, I get a topic rate of only `9.7Hz` (using `rostopic hz /base_laser_front/scan_raw`)

thus, I see that it hasn't been correct before (i.e. "doubled" `publish_frequency`), but it's not correct now either (i.e. `9.7Hz` != `12Hz`)

also, the "doubled" `publish_frequency` has been around for quite a while and I think our applications, e.g. navigation, rely on the topic rate of `25Hz`

however, setting the `publish_frequency` to `25Hz`, with this PR I only get a topic rate of `13.5Hz`
while with `ipa320/indigo_dev` the topic rate is `25Hz` again (wtf?)...it's even `25Hz` with `publish_frequency` set to `50Hz`...and it's `25Hz` with `publish_frequency` set to `5Hz`
thus, it seems that in current `ipa320/indigo_dev` the `publish_frequency` parameter has no actual influence on the publish rate at all...it's `25Hz` for any value!

as all this seems rather confusing, I would like to here some other maintainer's opinion (@ipa-fez @ipa-jba @floweisshardt @benmaidel) before merging!

BTW, what's the default/max publish_rate of the actual HW? is it configurable at all?

@mgruhler as you have been working with the sensor quite a bit, do you have anything to add to the discussion?

ipa-fez:
From a quick glance it looks like we just skip scans if the HW sends scans too fast, so if the hardware runs at 25 Hz and we request 20 Hz, we'll skip basically every other scan because the 1/20 ms haven't passed yet.

Maybe this should always publish at HW speed and topic_tools throttle can be used if lower HZ is required.

mgruhler:
I never looked into this parameter.

But I can share some insights (that are almost not documented in the Sick docus) about the possible frequencies and what influences those that I once added in https://github.com/ipa320/cob_driver/issues/338#issuecomment-311023596.

To summarize:
- Max Frequency is 25Hz for new protocol and 12.5Hz for old protocol scanners
- There are many settings in the Sick configuration software that further limits this

Also, when looking at the code, I guess that the parameter is effectively a noop, the way it is implemented right now:

```C++
if(ros::Time::now()-loop_rate_.now()>=ros::Duration(1./publish_frequency))
```
is actually calling `now()` on the `ros::Time loop_rate_` (which is probably the most misleading variable name i've ever encountered) object. I'm guessing this is returning the same as `ros::Time::now()`. Thus, this check is always false, leading to it not being used at all.
If it gets fixed though, @ipa-fez is right, we will just skip every other scan. And if you set it to something of which 25 is not a multiple of (i.e. 25/x, where x is an integer, basically everything except 12.5, 8,33333, 6,25, 5, ...), you will get scans with different spacings that could result in those strange frequencies that you see (as `rostopic hz` is actually averaging, if I'm correct).

I agree with @ipa-fez that the best way would be to remove this parameter completely and then throttle the topic if you actually need it at lower frequency.

fmessmer:
I'm going to close this unmerged given the discussion and with #377 being merged

kalectro:
Sorry for not replying, just saw everything today and like the solution

PR #374, opened 2018-06-08
round remaining_capacity

floweisshardt:
round the remaining_capacity from the bms fake node to 3 digits (same way as it is done by the real bms driver)


ipa-rmb (dismissed):

fmessmer:
but the rounding only takes place in the power_state_aggregator which is also used with the fake_bms (see https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/drivers/bms.launch)

I cant find any rounding in the real driver...

fmessmer:
I compared the precision of all fields to the real driver
in @floweisshardt's version, the precision in the diagnostics was still not correct. by rounding the member variables whenever they are set, the precision is correct in both publishers and diagnostics

PR #375, opened 2018-07-07
bulletproof mimic

fmessmer:
catching some error return values of the libvlc library that might give us a hint why sometimes mimics are not displayed or the player is not visible...


fmessmer:
running on paul-unity

PR #378, opened 2018-08-03
add auto reconnect when scanner is disconnected

mateuszcierpikowski:
Hi, 
I have a problem with disconnecting Sick S300 scanner. I noticed that your driver doesn't reconnect automatically when USB converter is detected by system again. So I added auto-reconnect feature.

### DIFF
--- a/cob_sick_s300/ros/src/cob_sick_s300.cpp
+++ b/cob_sick_s300/ros/src/cob_sick_s300.cpp
@@ -75,6 +75,7 @@ class NodeClass
 		unsigned int syncedSICKStamp;
 		bool syncedTimeReady;
 		bool debug_;
+		double communication_timeout;
 		ScannerSickS300 scanner_;
 		std_msgs::Bool inStandby_;
 
@@ -106,6 +107,9 @@ class NodeClass
 			nh.param("scan_cycle_time", scan_cycle_time, 0.040); //SICK-docu says S300 scans every 40ms
 
 			if(nh.hasParam("debug")) nh.param("debug", debug_, false);
+
+			if(!nh.hasParam("communication_timeout")) ROS_WARN("Used default parameter for communication timeout");
+			nh.param("communication_timeout", communication_timeout, 0.2);
 
 			try
 			{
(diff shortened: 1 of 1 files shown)


ipa-rmb:
@ipa-jba @ipa-owner-fez Can someone from the navigation team approve the addon?

fmessmer:
we never experienced disconnecting scanners...so I'm a bit anxious about adding something to this driver that we haven't missed so far..definitely needs to be well-tested on our platforms!
what's your setup? do you have lose cables?
do you use proper udev rules to map the scanners to specific `/dev/...` files so they do not change ar runtime...
I'd rather investigate why your scanners disconnect than adding a re-connect workaround

mateuszcierpikowski:
udev rules are configured properly. It's a problem with loose USB contacts. Sometimes scanners disconnect only for a second. After that scanner is properly detected by system but your driver doesn't publish scan data... You can test it if you detach scanner and connect it again.

fmessmer:
I'll try to find a slot to test this...a little patience please :wink:

fmessmer:
I had a quick look at this PR this morning

`indigo_dev`-behavior:
 - scanners plugged in, valid udev rules, start driver, scans published, diagnostics OK
 - unplug scanners, driver goes to "simulation mode", no scans published, diagnostics STALE
 - re-plug scanners, valid udev rules, driver stays in "simulation mode", no scans, diagnostics STALE

`feature/auto_reconnect`-behavior:
 - scanners plugged in, valid udev rules, start driver, scans published, diagnostics OK
 - unplug scanners, driver tries to re-connect, no scans, diagnostics ERROR (scanner not on port)
 - re-plug scanners, valid udev rules, scans published again, diagnostics OK again

thus, I'd say this PR is definitely an improvement

however, the part "re-plug scanners, valid udev rules", does not yet work for us that straight-forward:
we are using symlinked devices instead of fixed udev rules...thus, wen unplugging the scanners, those links become invalid....and do not become valid again after re-plugging the scanners....to make them valid again, our [udev script](https://github.com/mojin-robotics/cob_hardware_test/blob/indigo_dev/cob_devices_test/scripts/udev_s300.sh) has be executed once again....
I'd say, this is a problem on our side and we should modify our udev handling for the scanners...see mojin-robotics/setup_cob4#73


what's the opinion of the other maintainers here?
@ipa-fez @ipa-jba @ipa-rmb @floweisshardt @benmaidel

floweisshardt:
I'm ok with merging as the behavior doesn't change for us (we're typically not unplugging the scanners during operation!)

ipa-rmb:
I agree with @floweisshardt . The PR improves the stability of the driver in general while not making it worse for our robots which would have a problem either way after losing connection.

fmessmer:
thanks @mateuszcierpikowski

PR #379, opened 2018-08-06
fix "constexpr needed" compilation error

### LINKED ISSUE #370: cob_light fails to compile
When trying to build cob_light from source (in gentoo), it fails with the following trace:
https://paste.pound-python.org/show/6nPWo7E1mZ9jQydFplXq/

There seem to be two independent problems when compiling the package.
```
‘constexpr’ needed for in-class initialization of static data member ‘const double DistApproxMode::UPDATE_FREQ’ of non-integral type [-fpermissive]
static const float DIST_MIN = 0.3f;
```

```
/usr/include/boost/bind/bind.hpp:319:35: error: no match for call to ‘(boost::_mfi::mf1<void, IColorO, std::vector<color::rgba>&>) (IColorO*&, std::vector<color::rgba>)’
         unwrapper<F>::unwrap(f, 0)(a[base_type::a1_], a[base_type::a2_]);
```

Installed boost version is 1.65.

@fmessmer FYI

jabbenseth:
- fixes #370

### DIFF
--- a/cob_light/common/include/distApproxMode.h
+++ b/cob_light/common/include/distApproxMode.h
@@ -115,9 +115,9 @@ public:
 
     std::string getName(){ return std::string("DistApproxMode"); }
 
-    static const float DIST_MIN = 0.3f;
-    static const float DIST_MAX = 2.0f;
-    static const double UPDATE_FREQ = 50.0;
+    static constexpr float DIST_MIN = 0.3f;
+    static constexpr float DIST_MAX = 2.0f;
+    static constexpr double UPDATE_FREQ = 50.0;
 
 private:
     double _timer_inc;


fmessmer (changes requested):
apparently not "multi-distro-compatible"
do we need to introduce a `melodic-devel` branch`? or can it be made compatible?

also, please add melodic checks for travis here, too

fmessmer:
could you combine with #380 as a single PR for melodic migration?

jabbenseth:
sure. however with melodic this is still a WIP. There are some more errors (in cob_light)

jabbenseth:
is now in #380

PR #380, opened 2018-08-06
[Melodic] combined melodify pr

jabbenseth:
ref ipa320/care-o-bot#52

:warning: we will need to change the default branch for this repo :warning: 

---

with std::shared_ptr

(deprecation.. for melodic)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,12 @@ env:
     - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=debian
     - ROS_DISTRO=kinetic UPSTREAM_WORKSPACE=file
     - ROS_DISTRO=kinetic UPSTREAM_WORKSPACE=debian
+    - ROS_DISTRO=melodic UPSTREAM_WORKSPACE=file
+    - ROS_DISTRO=melodic UPSTREAM_WORKSPACE=debian
 matrix:
   allow_failures:
-    - env: ROS_DISTRO=indigo UPSTREAM_WORKSPACE=debian
-    - env: ROS_DISTRO=kinetic UPSTREAM_WORKSPACE=debian
+    - env: ROS_DISTRO=melodic UPSTREAM_WORKSPACE=file
+    - env: ROS_DISTRO=melodic UPSTREAM_WORKSPACE=debian
 install:
   - git clone https://github.com/ros-industrial/industrial_ci.git .ci_config
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
-sudo: required
-dist: trusty
 language: generic
+services:
+  - docker
+
 notifications:
   email:
     on_success: change
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,10 @@ env:
     - ROS_DISTRO=kinetic UPSTREAM_WORKSPACE=debian
     - ROS_DISTRO=melodic UPSTREAM_WORKSPACE=file
     - ROS_DISTRO=melodic UPSTREAM_WORKSPACE=debian
+matrix:
+  allow_failures:
+    - ROS_DISTRO=kinetic UPSTREAM_WORKSPACE=debian
+    - ROS_DISTRO=melodic UPSTREAM_WORKSPACE=debian
 install:
   - git clone https://github.com/ros-industrial/industrial_ci.git .ci_config
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ env:
   global:
     - ROS_REPO=ros
   matrix:
+    - ROS_DISTRO=kinetic UPSTREAM_WORKSPACE=file
+    - ROS_DISTRO=kinetic UPSTREAM_WORKSPACE=debian
     - ROS_DISTRO=melodic UPSTREAM_WORKSPACE=file
     - ROS_DISTRO=melodic UPSTREAM_WORKSPACE=debian
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,16 +9,8 @@ env:
   global:
     - ROS_REPO=ros
   matrix:
-    - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=file
-    - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=debian
-    - ROS_DISTRO=kinetic UPSTREAM_WORKSPACE=file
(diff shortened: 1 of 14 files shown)


fmessmer (changes requested):
apparently not "multi-distro-compatible"
do we need to introduce a `melodic-devel` branch`? or can it be made compatible?

also, please add melodic checks for travis here, too

jabbenseth:
Unfortunately we will need a melodic-devel branch. 
Some drivers use socketcan_interface where the pointers were renamed for example from `Interface::Ptr` to `InterfaceSharedPtr` and "retyped" from boost do std. I think it would be unreasonable effort to keep compatibility for kinetic(and earlier) and make it work for melodic

mathias-luedtke:
> Some drivers use socketcan_interface where the pointers were renamed

Actually, it should be API-compatible between kinetic and melodic.

jabbenseth:
> Actually, it should be API-compatible between kinetic and melodic.

Indeed. cob_drivers supports indigo and kinetic. in Listener for example the Listener::Ptr was marked deprecated but still used here https://github.com/ipa320/cob_driver/pull/380/files#diff-92aa77383a887ecfd66e061c1ef3e5e3R75

now with melodic the ::Ptr was removed and the introduced (since kinetic) …ConstSharedPtr is the only available option.

--> this is not an issue of API-compatibility between kinetic and melodic (which is indeed true) but the api compatibility between indigo and melodic.

Introducing a new kinetic branch while tackling issues produced by melodic seems counterintuitive to me even though we likely can support melodic and kinetic just by dropping indigo.

mathias-luedtke:
Indigo is almost EOL (~ 2 weeks left), but kinetic will be supported for the next two years..

fmessmer:
while introducing a `kinetic_dev` branch to deal with kinetic/melodic compatibility will result in some *.rosintall file trouble, it seems the correct way to do...
`indigo_dev` could still be used for kinetic - although it might not include the latest kinetic updates...
for the new branch `kinetic_dev` we should also bump the version to `0.7.X`

.travis.yml
```diff
@@ -9,16 +9,10 @@ env:
...
-    - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=file
-    - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=debian
-    - ROS_DISTRO=kinetic UPSTREAM_WORKSPACE=file
-    - ROS_DISTRO=kinetic UPSTREAM_WORKSPACE=debian
-matrix:
-  allow_failures:
-    - env: ROS_DISTRO=indigo UPSTREAM_WORKSPACE=debian
-    - env: ROS_DISTRO=kinetic UPSTREAM_WORKSPACE=debian
+    - ROS_DISTRO=melodic UPSTREAM_WORKSPACE=file
+    - ROS_DISTRO=melodic UPSTREAM_WORKSPACE=debian
```
fmessmer, line 16: if we merge this into a `kinetic_dev` branch (kinetic/melodic) we should include the kinetic jobs again and provide respective rosinstall file (if needed)

fmessmer:
on a `kinetic_dev` branch, we could even consider removing packages which we do not need anymore and we do not want to migrate to melodic

fmessmer:
on a side note:
something similar might need to be done for downstream repos...e.g. cob_robots is using `indigo_dev` branch which in fact has not been indigo-compatible for quite some time now...adding a new `kinetic_dev` branch in such cases will fix that dilemma

cob_elmo_homing/src/plugin.cpp
```diff
@@ -66,7 +66,7 @@ class ElmoMotor402 : public canopen::Motor402 {
     }
 
 public:
-    ElmoMotor402(const std::string &name, boost::shared_ptr<canopen::ObjectStorage> storage, const canopen::Settings &settings)
+    ElmoMotor402(const std::string &name, std::shared_ptr<canopen::ObjectStorage> storage, const canopen::Settings &settings)
```
mathias-luedtke: Use the pointer names here as well. Otherwise is does not work on kinetic.

jabbenseth:
The recent changes were 
a) fix the commented part by ipa-mdl
b) rebase on ipa320/indigo_dev
c) merge/cherry-pick #391

fmessmer:
So, this is going to be the PR where we collect everything needed to make `cob_driver` `kinetic`- and `melodic`-compatible?
I'd :+1: this

Then, once this PR passes Travis, we would create a new branch `kinetic_dev` and retarget the PR?
Then, `indigo_dev` is  `indigo`- and `kinetic`-compatible and `kinetic_dev` is `kinetic-` and `melodic`-compatible and will become the new default branch...

jabbenseth:
travis failed with rosdep issues and I found no option to restart --> close and open this PR

mathias-luedtke:
#391 does not work with kinetic. (at least out-of-the-box)
Enforcing C++11 might lead to ABI breaks.

cob_elmo_homing/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(cob_elmo_homing)
+add_compile_options(-std=c++11)
```
mathias-luedtke: Wrong file?

cob_elmo_homing/src/plugin.cpp
```diff
@@ -108,12 +108,14 @@ class ElmoMotor402 : public canopen::Motor402 {
...
-      virtual std::shared_ptr<canopen::MotorBase> allocate(const std::string& name,
-                                                           canopen::ObjectStorageSharedPtr storage,
-                                                           const canopen::Settings& settings)
+      virtual canopen::MotorBaseSharedPtr allocate(const std::string& name,
+                                                   canopen::ObjectStorageSharedPtr storage,
+                                                   const canopen::Settings& settings)
       {
-        return std::make_shared<ElmoMotor402>(name, storage, settings);
-        }
+        canopen::MotorBaseSharedPtr mbsp;
```
mathias-luedtke: please use the `ROSCANOPEN_MAKE_SHARED` macro

cob_elmo_homing/CMakeLists.txt
mathias-luedtke: And remove it here, I cannot guarantee that it works properly with C++1 mixed in for kinetic

fmessmer:
I changed the target branch for this PR to **new branch** `kinetic_dev` - as it is no longer indigo compatible

fmessmer (approved):

PR #383, opened 2018-09-12
add laser scan densifier

fmessmer:
a node adding additional sensor points to a LaserScan topic...

interpolating between two data points does not really work using `/scan_unified` as input as this node overwrites some ranges with plain 0...resulting in long fake points along a virtual ray

inserting data points works looks better:
screenshot from 2018-09-12 17-48-33

### DIFF
--- a/laser_scan_densifier/include/laser_scan_densifier/laser_scan_densifier_nodelet.h
+++ b/laser_scan_densifier/include/laser_scan_densifier/laser_scan_densifier_nodelet.h
@@ -58,7 +58,7 @@ class LaserScanDensifierNodelet : public nodelet::Nodelet
     virtual void onInit();
 
   private:
-    LaserScanDensifier * laser_scan_densifier_;  // FIXME: change to smart pointer
+    boost::shared_ptr< LaserScanDensifier > laser_scan_densifier_;
 };
 
 } //namespace scan_tools
--- a/laser_scan_densifier/include/laser_scan_densifier/laser_scan_densifier_nodelet.h
+++ b/laser_scan_densifier/include/laser_scan_densifier/laser_scan_densifier_nodelet.h
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2017 Fraunhofer Institute for Manufacturing Engineering and Automation (IPA)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


laser_scan_densifier/src/laser_scan_densifier.cpp
```diff
@@ -0,0 +1,110 @@
...
+  scan_dense->angle_min       = scan_msg->angle_min;
+  scan_dense->angle_max       = scan_msg->angle_max;
+  scan_dense->angle_increment = scan_msg->angle_increment / step_;
+  scan_dense->time_increment  = scan_msg->time_increment;
+  scan_dense->scan_time       = scan_msg->scan_time;
+
+  scan_dense->ranges.clear();
+  scan_dense->intensities.clear();
+
+  for (unsigned int i = 0; i < scan_msg->ranges.size()-1; i++)
```
ipa-rmb: 1. even though usingned int is correct, size_t is the best choice for the type of i, since that is the type returned by .size()

2. Why i< .size()-1 (don't you miss the last sensor value by using < and -1 at the same time)? Why is the last element just inserted once after the loop?

laser_scan_densifier/include/laser_scan_densifier/laser_scan_densifier_nodelet.h
```diff
@@ -0,0 +1,66 @@
...
+
+namespace scan_tools {
+
+class LaserScanDensifierNodelet : public nodelet::Nodelet
+{
+  public:
+    virtual void onInit();
+
+  private:
+    LaserScanDensifier * laser_scan_densifier_;  // FIXME: change to smart pointer
```
ipa-rmb: Yes, please use a smart pointer

laser_scan_densifier/src/laser_scan_densifier.cpp
```diff
@@ -0,0 +1,110 @@
...
+  {
+    scan_dense->ranges.insert(scan_dense->ranges.end(), step_, scan_msg->ranges[i]);
```
ipa-rmb: Would it make more sense to interpolate all intermediate values by the average, median or Gaussian weighted average from the neighboring measurements from left and right instead of just copying a value to the right?

ipa-rmb (changes requested):
Looks good, I just added a few comments.

laser_scan_densifier/src/laser_scan_densifier.cpp
fmessmer: if I'd do `size()`, I would change the `angle_max` of the scan...I don't want to add data points "right of" the original angle_max data point

laser_scan_densifier/src/laser_scan_densifier.cpp
fmessmer: I now support both modes...in our usecase...we apply this node onto processed data...where interpolation introduces false data points we don't want...

PR #384, opened 2018-10-26
[WIP] Added Nuvation NUV300 BMS driver

ipa-rmb:
This driver allows to read current and voltage measurements, and maybe even state of charge measurements from the Nuvation NUV300 battery management system.

The delivered config file is still under development, the remainder of the software is stable and well-tested.

### DIFF
--- a/nuvation_nuv300_bms_driver/CMakeLists.txt
+++ b/nuvation_nuv300_bms_driver/CMakeLists.txt
@@ -16,8 +16,7 @@ find_package(catkin REQUIRED COMPONENTS
 	${catkin_BUILD_PACKAGES}
 )
 
-
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake")
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${PROJECT_SOURCE_DIR}/cmake")
 
 find_package(Modbus)
 
--- a/nuvation_nuv300_bms_driver/CMakeLists.txt
+++ b/nuvation_nuv300_bms_driver/CMakeLists.txt
@@ -1,0 +1,70 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(nuvation_nuv300_bms_driver)
+
+# add_compile_options(-std=c++11)
+
+set(catkin_RUN_PACKAGES
+	cob_msgs
+	roscpp
+)
+
+set(catkin_BUILD_PACKAGES
+	${catkin_RUN_PACKAGES}
+)
+
+find_package(catkin REQUIRED COMPONENTS
+	${catkin_BUILD_PACKAGES}
+)
+
+
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake")
+
+find_package(Modbus)
+
+
+###################################
+## catkin specific configuration ##
+###################################
+catkin_package(
+#	INCLUDE_DIRS ros/include
+#	LIBRARIES nuvation_nuv300_bms_driver
+	CATKIN_DEPENDS
+		${catkin_RUN_PACKAGES}
+	DEPENDS
+		Modbus
+)
+
+###########
+## Build ##
+###########
+
+include_directories(
+	#ros/include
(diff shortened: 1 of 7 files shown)


nuvation_nuv300_bms_driver/ros/src/nuvation_nuv300_bms_driver_node.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+class Nuv300BMS
+{
+public:
+	Nuv300BMS(ros::NodeHandle& nh)
+	: node_handle_(nh), sequence_counter_(0)
+	{
+		// parameters
+		ros::NodeHandle pnh("~");
+		std::cout << "\n========== Nuv300BMS Parameters ==========\n";
```
fmessmer, line 30: use ROS logging, and not `cout` and also not `fprintf`

nuvation_nuv300_bms_driver/ros/src/nuvation_nuv300_bms_driver_node.cpp
```diff
@@ -0,0 +1,132 @@
...
+		if (modbus_read_registers(mb, 40132, 1, tab_reg) == -1)
+			fprintf(stderr, "%s\n", modbus_strerror(errno));
+		const double bcurrent_sf = pow(10,(double)((int16_t*)tab_reg)[0]);
+		
+		std::cout << "soc_sf=" << soc_sf << "   vol_sf=" << vol_sf << "   bcurrent_sf=" << bcurrent_sf << std::endl;
+
+		ros::Rate loop_rate(loop_rate_);
+		while(ros::ok())
+		{
+			// From: https://www.nuvationenergy.com/technical-resources
```
fmessmer, line 63: this `while` body could be separated as a ROS Timer callback

nuvation_nuv300_bms_driver/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+	${catkin_RUN_PACKAGES}
+)
+
+find_package(catkin REQUIRED COMPONENTS
+	${catkin_BUILD_PACKAGES}
+)
+
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${PROJECT_SOURCE_DIR}/cmake")
+
+find_package(Modbus)
```
fmessmer, line 21: why does Modbus needs to be found like this?

fmessmer (changes requested):
is this really a **cob** driver?
in what platform is this used?
I'd suggest to put this in a separate repo

why not just call the repo/package `nuv_bms_driver` and keep the full model only in the node name?

nuvation_nuv300_bms_driver/CMakeLists.txt
ipa-rmb: Because it was not found else, I needed to create file nuvation_nuv300_bms_driver/cmake/FindModbus.cmake to make it work. Is there another option?

libmodbus-dev is referenced in package.xml like you would refere to libopencv-dev or boost. It is installed with resdep install. So I imitated the opencv or boost inclusion processes here.

ipa-rmb:
The BMS is used with raw 3 and 4 platforms. I placed it here because all other variants of BMS also appear here and to me it made sense to have the same things (BMS drivers) at the same place.

What alternative repository would you suggest?

jabbenseth:
Sorry for stepping in: creating a new public nuv_bms_driver and (quickly) releasing it would be my preferred way

ipa-rmb:
@ipa-jba You mean creating a new public repository termed nuv_bms_driver containing the nuvation_nuv300_bms_driver package, correct? I can manage that but would prefer you to make the release since I am on vacation now.

jabbenseth:
I created a new package (https://github.com/ipa320/nuv_bms_driver) and adapted this PR: https://github.com/ipa320/nuv_bms_driver/pull/2 @ipa-rmb  @fmessmer 

I found no nicer "find modbus" solution either @fmessmer 
For the future I think adding a driver for the provided CAN interface might be worth investigating.

A Release has to wait a little longer since cob_msgs is not released for melodic (powerstate message is needed)

fmessmer:
now lives in new repo: https://github.com/ipa320/nuv_bms_driver

PR #385, opened 2018-11-21
initialize ranges with max_range rather than zero

fmessmer:
situation before:
 - `cob_sick_s300` driver publishes some ranges as `29.959999084472656` - most likely due to reflections e.g. from railings or glass fronts
 - these data points are removed during the projection in `cob_scan_unifer` - because they are `>max_range` (max_range=29.5)
 - the `unified_scan.ranges` are initialized with `0.0`
 - for those data points removed during projection, `scan_unified` will have a range of `0.0` which later gets removed during `laser_zone_filter` resulting in "holes" in the scan
 - along the rays within those "holes" the costmap cannot clear the obstacle layer

situation now:
 - now scan_unifier behaves as the driver: fake data points are set to `max_range`

this significantly reduces "trails" in the costmap for the scan unified obstacle layer

I already verified the behavior on `cob4-22` in 25h köln....and it looks much better

contributes to https://github.com/mojin-robotics/cob4/issues/888#issuecomment-439805060


HannesBachter (changes requested):
Please add a comment (in the code) on why you changed the default distance to `max_range` in order to avoid confusion in a future investigation

PR #386, opened 2018-11-27
[cob_phidgets] srv callback check if digital sensor is already at desired state

benmaidel:
let the srv call succeed if the digital sensor is already at the desired state


fmessmer (changes requested):
indentation has changed from tabs to spaces...please stick with tabs

fmessmer (approved):

PR #387, opened 2018-12-10
make scan_unifier work with non static tf transforms to laser frame

cme-research:
Issue: Using scan_unifier with non static tf transforms skips many scans because TF Lookup would require extrapolation into the future.

Source:
http://wiki.ros.org/laser_geometry --> 2.2 High fidelity projection


ipa-rmb (dismissed):
Looks reasonable to me. Could @ipa-jba or @ipa-owner-fez please confirm and merge?

fmessmer:
I would like to run tests on a robot before merging...
...sorry I didnt already do them...

ipa-rmb:
Of course, please test it with a real robot first.

cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -212,7 +212,7 @@ bool ScanUnifierNode::unifyLaserScans(std::vector<sensor_msgs::LaserScan::ConstP
       try
       {
         if (!listener_.waitForTransform(frame_, current_scans.at(i)->header.frame_id,
-                                        current_scans.at(i)->header.stamp, ros::Duration(1.0)))
+                                        current_scans.at(i)->header.stamp + ros::Duration().fromSec((current_scans.at(i)->ranges.size()-1)*current_scans.at(i)->time_increment), ros::Duration(1.0)))
```
fmessmer: why did you use `ranges.size()-1`, instead of `ranges.size()`?
your source states that you need to wait for the last scan, i.e. wait `time_increment` for every range value...with `-1` you are missing one `time_increment`...

fmessmer (changes requested):
tests on robot HW still pending...

cob_scan_unifier/src/scan_unifier_node.cpp
cme-research: Alright changed as requested.
```suggestion
                                        current_scans.at(i)->header.stamp + ros::Duration().fromSec(current_scans.at(i)->ranges.size()*current_scans.at(i)->time_increment), ros::Duration(1.0)))
```

PR #388, opened 2019-02-18
choose default and blinking mimic

HannesBachter:
enables choosing the default and blinking mimic for cob-mimic 
e.g. start in roslaunch as follows (`default` needs to be a string, `blinking` needs to be a list of strings):
```
<arg name="default"/>
<arg name="blinking"/>
<node pkg="cob_mimic" type="mimic" name="mimic" cwd="node" respawn="false" output="screen">
		<param name="default" value="$(arg default)"/>
                <rosparam param="blinking" subst_value="True">$(arg blinking)</rosparam> 
</node>
```


fmessmer:
closing in favor of #410

PR #389, opened 2019-03-26
Initalise TF2 before using it. Otherwise runtime error in current kin…

destogl:
…etic

 - fixes problem with not initialized tf2_Buffer. Crashes with current kinetic.

### DIFF
--- a/cob_scan_unifier/src/scan_unifier_node.cpp
+++ b/cob_scan_unifier/src/scan_unifier_node.cpp
@@ -26,6 +26,10 @@ ScanUnifierNode::ScanUnifierNode()
   nh_ = ros::NodeHandle();
   pnh_ = ros::NodeHandle("~");
 
+  // Init TF-Buffer
+  p_tfBuffer = new tf2_ros::Buffer();
+  p_tfListener = new tf2_ros::TransformListener(*p_tfBuffer, true);
+
   // Publisher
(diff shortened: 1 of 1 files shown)


cob_scan_unifier/src/scan_unifier_node.cpp
```diff
@@ -26,6 +26,10 @@ ScanUnifierNode::ScanUnifierNode()
   nh_ = ros::NodeHandle();
   pnh_ = ros::NodeHandle("~");
 
+  // Init TF-Buffer
+  p_tfBuffer = new tf2_ros::Buffer();
+  p_tfListener = new tf2_ros::TransformListener(*p_tfBuffer, true);
```
fmessmer, line 31: but this node does not use `tf2`...so what kind of  errors are you experience?
(CI fails as this node does not have those members you are initializing)

fmessmer (changes requested):
do not merge

fmessmer:
@destogl maybe this is what you want to do: https://github.com/ipa320/cob_control/pull/202
@ipa-jba FYI

cob_scan_unifier/src/scan_unifier_node.cpp
destogl: I do not get unified scans. As they are not transformed since tf is not initalized... would it be solution to move to tf2 or should I implement as in ipa320/cob_control#202 ?

fmessmer:
not planned

PR #393, opened 2019-06-05
add int bms parameter types

fmessmer:
the BMS CAN protocol contains a few variables that are neither Float nor Bool, so adding Int/UInt seems helpfull....

currently testing an extended config on our robots - with more/full diagnostic infos...

### DIFF
--- a/cob_bms_driver/CMakeLists.txt
+++ b/cob_bms_driver/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(cob_bms_driver)
 
+add_compile_options(-std=c++11)
+
 find_package(catkin REQUIRED COMPONENTS
   diagnostic_msgs
   diagnostic_updater
--- a/cob_bms_driver/CMakeLists.txt
+++ b/cob_bms_driver/CMakeLists.txt
@@ -1,8 +1,6 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(cob_bms_driver)
 
-add_compile_options(-std=c++11)
-
 find_package(catkin REQUIRED COMPONENTS
   diagnostic_msgs
   diagnostic_updater
--- a/cob_bms_driver/src/cob_bms_driver_node.cpp
+++ b/cob_bms_driver/src/cob_bms_driver_node.cpp
@@ -77,7 +77,6 @@ template<typename T> struct TypedBmsParameter : BmsParameter {
     virtual void advertise(ros::NodeHandle &nh, const std::string &topic){
         publisher = nh.advertise<T> (topic, 1, true);
     }
-
 };
 
 struct FloatBmsParameter : TypedBmsParameter<std_msgs::Float64> {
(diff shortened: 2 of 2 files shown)


fmessmer:
@ipa-mdl 
do you have any objection/opinion on this change?

it gives me Int/UInt numbers in the diagnostics instead of Float for fields which express a counter value
Int/Uint vs. Float is decided based on the existence of the `factor` entry in the config

fmessmer:
however, I'm still having trouble do correctly decode the following CAN IDs (from the docu):
```
0x010B      Seriennummer BMZ lesen       4 Bytes                  0x07.5B.CD.15
                                         unsigned long            = Ser.Nr.: 123456789
0x010C      Seriennummer FSM lesen       4 Bytes
0x010D      Herst. Datum lesen           6 Bytes                  0x00.63.00.0C.00.1F
             Jahr                           Byte 0                = Datum: 31.12.99
             Jahr                           Byte 1
             Monat                          Byte 2
             Monat                          Byte 3
             Tag                            Byte 4
             Tag                            Byte 5
```

but this PR gives me the following diagnostics:
```
serial_number_bmz: 0
serial_number_fsm: 376
manufacturing_date_year_1: 0
manufacturing_date_year_2: 127
manufacturing_date_month_1: 0
manufacturing_date_month_2: 15
manufacturing_date_day_1: 0
manufacturing_date_day_2: 31
```
where `manufacturing_date_year_2: 127` and `manufacturing_date_month_2: 15` doesn't really make sense

config for this is:
```
  - id: 0x0B # Serial BMZ
    fields:
      - name: "serial_number_bmz"
        offset: 0
        len: 4
        is_signed: false

  - id: 0x0C # Serial FSM
    fields:
      - name: "serial_number_fsm"
        offset: 0
        len: 4
        is_signed: false

  - id: 0x0D # Manufacturing Date
    fields:
      - name: "manufacturing_date_year_1"
        offset: 0
        len: 1
        is_signed: false
      - name: "manufacturing_date_year_2"
        offset: 1
        len: 1
        is_signed: false
      - name: "manufacturing_date_month_1"
        offset: 2
        len: 1
        is_signed: false
      - name: "manufacturing_date_month_2"
        offset: 3
        len: 1
        is_signed: false
      - name: "manufacturing_date_day_1"
        offset: 4
        len: 1
        is_signed: false
      - name: "manufacturing_date_day_2"
        offset: 5
        len: 1
        is_signed: false
```

fmessmer:
never mind my https://github.com/ipa320/cob_driver/pull/393#issuecomment-498955172
apparently there is crappy data on the CAN-Bus for those fields in the first place and the extension in this PR does the right thing

@ipa-bdw read the respective objects with PCAN-Viewer and got the following results:
```
010A: 01 00 01 02
010B: 00 00 00 00
010C: 00 00 01 78
010D: 00 7F 00 0F 00 1F
```
given this, the diagnostics above are converting the data correctly

cob_bms_driver/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(cob_bms_driver)
 
+add_compile_options(-std=c++11)
```
mathias-luedtke: Don't do this, it will open Pandora's box.

fmessmer: using `boost::lexical_cast` instead now...sufficient for your approval?

mathias-luedtke (approved):

cob_bms_driver/CMakeLists.txt
jabbenseth: can you elaborate why this is not the way to specify the usage of c++11(14, 17)..?

fmessmer: @ipa-mdl suggests not to use c++11 for kinetic at all - abi-compatibility...

PR #394, opened 2019-07-02
[cob_mimic] publish diagnostics

HannesBachter:
...and use multithreaded spinner for diagostics not to go stale during service calls
fixes https://github.com/mojin-robotics/cob4/issues/1143

### DIFF
--- a/cob_mimic/CMakeLists.txt
+++ b/cob_mimic/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(cob_mimic)
 
+add_compile_options(-std=c++11)
+
 find_package(catkin REQUIRED COMPONENTS actionlib actionlib_msgs message_generation roscpp roslib)
 find_package(PkgConfig REQUIRED)
 find_package(Boost REQUIRED COMPONENTS filesystem thread)
--- a/cob_mimic/CMakeLists.txt
+++ b/cob_mimic/CMakeLists.txt
@@ -1,8 +1,6 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(cob_mimic)
 
-add_compile_options(-std=c++11)
-
 find_package(catkin REQUIRED COMPONENTS actionlib actionlib_msgs message_generation roscpp roslib)
 find_package(PkgConfig REQUIRED)
 find_package(Boost REQUIRED COMPONENTS filesystem thread)
--- a/cob_mimic/src/mimic_node.cpp
+++ b/cob_mimic/src/mimic_node.cpp
@@ -71,7 +71,7 @@ public:
 
         diagnostic_updater_.setHardwareID("none");
         diagnostic_updater_.add("mimic", this, &Mimic::produce_diagnostics);
-        diagnostic_timer_ = nh_.createTimer(ros::Duration(1.0), &Mimic::diagnostics_timer_cb, this);
+        diagnostic_thread_ = boost::thread(&Mimic::diagnostics_timer_thread, this);
 
         int_dist_ = boost::random::uniform_int_distribution<>(0,static_cast<int>(random_mimics_.size())-1);
 
@@ -113,7 +113,7 @@ private:
     bool service_active_;
     std::string active_mimic_;
     diagnostic_updater::Updater diagnostic_updater_;
-    ros::Timer diagnostic_timer_;
+    boost::thread diagnostic_thread_;
 
     libvlc_instance_t* vlc_inst_;
     libvlc_media_player_t* vlc_player_;
@@ -346,9 +346,13 @@ private:
         return true;
     }
 
-    void diagnostics_timer_cb(const ros::TimerEvent&)
+    void diagnostics_timer_thread()
     {
-        diagnostic_updater_.update();
+        while(ros::ok())
+        {
+          ros::Duration(1.0).sleep();
+          diagnostic_updater_.update();
+        }
     }
 
     void produce_diagnostics(diagnostic_updater::DiagnosticStatusWrapper &stat)
@@ -365,7 +369,6 @@ int main(int argc, char** argv)
 {
     ros::init(argc, argv, "mimic");
 
-    ros::AsyncSpinner spinner(4);
     Mimic mimic;
     if(!mimic.init())
     {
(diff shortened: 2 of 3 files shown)


HannesBachter:
tested on cob4-19 :heavy_check_mark:

cob_mimic/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(cob_mimic)
 
+add_compile_options(-std=c++11)
```
fmessmer: for the same reason as in https://github.com/ipa320/cob_driver/pull/393#discussion_r290780814

cob_mimic/src/mimic_node.cpp
```diff
@@ -174,18 +185,21 @@ class Mimic
...
 
         if(req.mimic != "falling_asleep" && req.mimic != "sleeping")
             blinking_timer_ = nh_.createTimer(ros::Duration(real_dist_(gen_)), &Mimic::blinking_cb, this, true);
+        service_active_ = false;
         return true;
     }
 
     bool set_mimic(std::string mimic, int repeat, float speed, bool blocking=true)
     {
+        active_mimic_= "Mimic: "+ mimic +", repeat: "+ std::to_string(repeat) +", speed: "+ std::to_string(speed) +", blocking: "+ std::to_string(blocking);
```
fmessmer: solve it as in https://github.com/ipa320/cob_driver/pull/393/files#diff-942a6ce5dfeab4525f7e1b076370b5ecR99 (without needing `add_compile_options(-std=c++11)`)

cob_mimic/src/mimic_node.cpp
```diff
@@ -202,6 +216,7 @@ class Mimic
             {
                 ROS_ERROR("File not found: %s", filename.c_str());
                 mutex_.unlock();
+                active_mimic_ = "None";
```
fmessmer, line 221: should this go before the `unlock`?

cob_mimic/src/mimic_node.cpp
```diff
@@ -231,6 +246,7 @@ class Mimic
             {
                 ROS_ERROR("failed to create media for filepath %s", filename.c_str());
                 mutex_.unlock();
+                active_mimic_ = "None";
```
fmessmer, line 251: should this go before the `unlock`?

cob_mimic/src/mimic_node.cpp
```diff
@@ -242,6 +258,7 @@ class Mimic
             {
                 ROS_ERROR("failed to play");
                 mutex_.unlock();
+                active_mimic_ = "None";
```
fmessmer, line 263: should this go before the `unlock`?

cob_mimic/src/mimic_node.cpp
```diff
@@ -254,12 +271,14 @@ class Mimic
                 {
                     ROS_WARN("mimic %s preempted", mimic.c_str());
                     mutex_.unlock();
+                    active_mimic_ = "None";
```
fmessmer, line 276: should this go before the `unlock`?

cob_mimic/src/mimic_node.cpp
```diff
@@ -254,12 +271,14 @@ class Mimic
...
                     return false;
                 }
             }
             repeat --;
         }
         mutex_.unlock();
+        active_mimic_ = "None";
```
fmessmer, line 283: should this go before the `unlock`?

fmessmer (changes requested):

fmessmer (changes requested):
also travis fails....

cob_mimic/src/mimic_node.cpp
HannesBachter: done (also for the others)

HannesBachter:
running on cob4-19 for a while now without any problems

cob_mimic/src/mimic_node.cpp
```diff
@@ -75,12 +82,11 @@ class Mimic
             "-q",
             "--no-osd",
             "-L",
-            "--one-instance",
+            "--no-one-instance",
             "--playlist-enqueue",
             "--no-video-title-show",
             "--no-skip-frames",
             "--no-audio",
-            "--vout=glx,none"
```
fmessmer: what do these arg changes do?

cob_mimic/src/mimic_node.cpp
```diff
@@ -324,6 +345,23 @@ class Mimic
...
+          diagnostic_updater_.update();
+        }
+    }
+
+    void produce_diagnostics(diagnostic_updater::DiagnosticStatusWrapper &stat)
+    {
+        stat.summary(diagnostic_msgs::DiagnosticStatus::OK, "Mimic running");
+        stat.add("Action is active", action_active_);
+        stat.add("Service is active", service_active_);
+        stat.add("Current goal", active_mimic_);
```
fmessmer: I would replace `Current goal` with `Active mimic` then...as it could also be the default mimic or the mimic requested by service...

fmessmer (changes requested):
minor....

cob_mimic/src/mimic_node.cpp
benmaidel: These arg changes where necessary to get it running under melodic. Because the node is running stable on melodic with this changes, we kept them here as well. 
- one-instance: Allow only one running instance
- vout: default video output mode => glx,none = opengl via glx
  if it is not defined, vlc will autopick the best option

fmessmer: thanks

 - so now we allow more than one instances - the flag is now set to `--no-one-intance`??? in what case do we need this?
 - is the "best option" in our setup something different then `glx,none`? does the auto-select feature improve performance by chance?

benmaidel: > * so now we allow more than one instances - the flag is now set to `--no-one-intance`??? in what case do we need this?

I think, that `one-instance` is the right option we need, because we do not like to open another window for a new video. Switching to `no-one-instance` was only done, because `one-instance` was not working on melodic. @HannesBachter  can you test it with the original option?



> * is the "best option" in our setup something different then `glx,none`? does the auto-select feature improve performance by chance?

`glx,none` was the only option that worked on older robots. We should either keep the option for now or test if it works on all our robots without the explicit vout option

HannesBachter: tested with the old parameters for vlc and did not crash then, too

PR #397, opened 2019-07-18
[mimic] feature/reduce load - compress videos

LoyVanBeek:
- fixes https://github.com/mojin-robotics/cob4/issues/1193

Decoding highly compressed .mp4 files requires a lot of CPU. This PR introduces a parameter that will decompress the files either at init or during a mimic. 
The latter option will make that a *next* use of that mimic is faster.

### DIFF
--- a/cob_mimic/CMakeLists.txt
+++ b/cob_mimic/CMakeLists.txt
@@ -1,9 +1,7 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(cob_mimic)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-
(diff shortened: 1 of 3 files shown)


floweisshardt:
not needed anymore for mimic videos due to https://github.com/ipa320/cob_driver/pull/398.

do we still want to merge this then?

HannesBachter:
if the first start of a mimic is still as fast as it was before (and subsequent loadings are quicker), I would merge (as we're only improving)

LoyVanBeek:
I wouldn't merge it. Compressing files is not the responsibility of `cob_mimic`, it feels like a bit of a hack and it kind of is... The alternative is much cleaner and simpler in that regard

fmessmer (changes requested):
yes, I wouldn't merge as well..but we could keep it open in case it pops up in the future...

PR #406, opened 2019-12-10
[cob_light] removed unused Boost signals dependency

lopsided98:
`cob_light` depends on Boost signals2, which is a header only library that does not need to be found using `find_package`. The `CMakeLists.txt` tries to find Boost signals (v1). This library was removed in Boost 1.69, causing this package to fail to build.

### DIFF
--- a/cob_light/CMakeLists.txt
+++ b/cob_light/CMakeLists.txt
@@ -5,7 +5,7 @@ add_compile_options(-std=c++11)
 
 find_package(catkin REQUIRED COMPONENTS actionlib_msgs actionlib diagnostic_msgs message_generation roscpp sensor_msgs std_msgs visualization_msgs)
 
-find_package(Boost REQUIRED COMPONENTS signals thread)
+find_package(Boost REQUIRED COMPONENTS thread)
 
 ### Message Generation ###
 add_message_files(


fmessmer:
sounds reasonable

just for my understanding:
in what architecture/distro/setup do you get build errors?
can you post a log?

lopsided98:
I am using [Nix](https://github.com/lopsided98/nix-ros-overlay), but this problem will occur on any distribution with Boost >=1.69. This is the error message:
```
CMake Error at /nix/store/0rwxndjblrj8ashnfv0g9sxcnv2mdiw3-cmake-3.15.4/share/cmake-3.15/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find Boost (missing: signals) (found version "1.69.0")
Call Stack (most recent call first):
  /nix/store/0rwxndjblrj8ashnfv0g9sxcnv2mdiw3-cmake-3.15.4/share/cmake-3.15/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /nix/store/0rwxndjblrj8ashnfv0g9sxcnv2mdiw3-cmake-3.15.4/share/cmake-3.15/Modules/FindBoost.cmake:2161 (find_package_handle_standard_args)
  CMakeLists.txt:8 (find_package)


-- Configuring incomplete, errors occurred!
```

Many ROS packages have this same problem: https://github.com/ros-planning/navigation/pull/945, https://github.com/ros/geometry2/pull/354, https://github.com/ros/geometry/pull/180, https://github.com/ros-planning/moveit/pull/1353, https://github.com/ros/ros_comm/pull/1580, https://github.com/PR2/pr2_common_actions/pull/39, https://github.com/ros-perception/calibration/pull/43

fmessmer (approved):
thx for the patch and the explanation....I took care of other `cob_*` repos with the same problem....

PR #408, opened 2020-01-05
[travis] ci updates

fmessmer:


fmessmer:
speed-up:
 - before:  Ran for 15 min 37 sec 
 - after: Ran for 9 min 29 sec

PR #411, opened 2020-02-20
[cob_light] added new modes

JeroenKempen:
* added mode for a turnsignal to the left
* added mode for a turnsingal to the right
* optimized datatypes in different places

For the turnsignal to use simple state the color in witch the turn signal should be blinking and the frequency.

### DIFF
--- a/cob_light/common/include/signalRightMode.h
+++ b/cob_light/common/include/signalRightMode.h
@@ -1,0 +1,75 @@
+#ifndef SIGNAL_RIGHT_H
+#define SIGNAL_RIGHT_H
+
+#include <vector>
+
+#include "mode.h"
+
+class SignalRight : public Mode
+{
+public:
+  SignalRight(color::rgba color, uint16_t nLeds, uint8_t priority = 0, double freq = 0.25, uint8_t pulses = 0,
+              double timeout = 0)
+    : Mode(priority, freq, pulses, timeout), _num_leds(nLeds), _toggle(false), _timer_inc(0), leds_margin(5)
+  {
+    _color = color;
+    if (_pulses != 0)
+    {
+      _pulses *= 2;
+      _pulses += 1;
+    }
+    _inc = (1. / UPDATE_RATE_HZ) * _freq;
+  }
+
+  virtual void execute()
+  {
+    std::vector<color::rgba> cols;
+    color::rgba turnedOff;
+    turnedOff.a = 0;
+
+    cols.push_back(turnedOff);
+    cols.push_back(turnedOff);
+    cols.push_back(turnedOff);
+    cols.push_back(turnedOff);
+    cols.push_back(turnedOff);
+
+    for (uint16_t i = leds_margin; i < (_num_leds / 2) - leds_margin; ++i)
+    {
+      cols.push_back(_color);
+    }
+    for (uint16_t i = (_num_leds / 2) - leds_margin; i < _num_leds; ++i)
+    {
+      cols.push_back(turnedOff);
+    }
+
+    if (_timer_inc >= 1.0)
+    {
+      for (color::rgba& color : cols)
+      {
+        color.a = color.a * (int)_toggle;
+      }
+      _pulsed++;
+      _toggle = !_toggle;
+      m_sigColorsReady(cols);
+      _timer_inc = 0.0;
+    }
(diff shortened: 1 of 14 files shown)


fmessmer:
thanks for the contribution
this is related to https://github.com/mojin-robotics/cob4/issues/1414

@benmaidel @LoyVanBeek 
please review

cob_light/common/include/signalRightMode.h
```diff
@@ -0,0 +1,75 @@
...
+    }
+    else
+    {
+      _timer_inc += _inc;
+    }
+  }
+
+  std::string getName()
+  {
+    return std::string("SignalRIght");
```
LoyVanBeek, line 64: Small typo.

cob_light/common/include/signalRightMode.h
```diff
@@ -0,0 +1,75 @@
...
+    cols.push_back(turnedOff);
+    cols.push_back(turnedOff);
+    cols.push_back(turnedOff);
+    cols.push_back(turnedOff);
+
+    for (uint16_t i = leds_margin; i < (_num_leds / 2) - leds_margin; ++i)
+    {
+      cols.push_back(_color);
+    }
+    for (uint16_t i = (_num_leds / 2) - leds_margin; i < _num_leds; ++i)
```
LoyVanBeek, line 40: I think this is the only difference between the left and right modes, correct? 
Is there really no way to have these both derive from a common class (eg. `DirectionIndicatorMode` or something better) and have a `direction` variable set to either left/right that makes the quintessential difference between these modes?

cob_light/ros/src/cob_light.cpp
```diff
@@ -463,16 +493,16 @@ class LightControl
...
+  ros::init(argc, argv, "cob_light");
+//  signal(SIGINT, sigIntHandler);
 
   // Override XMLRPC shutdown
-  ros::XMLRPCManager::instance()->unbind("shutdown");
-  ros::XMLRPCManager::instance()->bind("shutdown", shutdownCallback);
+//  ros::XMLRPCManager::instance()->unbind("shutdown");
+//  ros::XMLRPCManager::instance()->bind("shutdown", shutdownCallback);
 
   // create LightControl instance
```
LoyVanBeek, line 503: Why has this been removed?

cob_light/ros/src/stageprofi.cpp
```diff
@@ -54,31 +54,35 @@ bool StageProfi::init()
...
+  // color_tmp.r *= color.a;
+  // color_tmp.g *= color.a;
+  // color_tmp.b *= color.a;
 
-  color_tmp.r = fabs(color_tmp.r * 255);
-  color_tmp.g = fabs(color_tmp.g * 255);
-  color_tmp.b = fabs(color_tmp.b * 255);
+  // color_tmp.r = fabs(color_tmp.r * 255);
+  // color_tmp.g = fabs(color_tmp.g * 255);
+  // color_tmp.b = fabs(color_tmp.b * 255);
```
LoyVanBeek, line 68: Why has this been removed?

cob_light/ros/src/stageprofi.cpp
```diff
@@ -93,6 +97,7 @@ void StageProfi::setColor(color::rgba color)
 
 void StageProfi::setColorMulti(std::vector<color::rgba> &colors)
 {
+  ROS_WARN(__PRETTY_FUNCTION__);
```
LoyVanBeek, line 100: Should be ROS_DEBUG at best I guess, if still needed at all

LoyVanBeek (changes requested):
Thanks for the contribution! 

The PR adds a few handy modes but also a lot of reformatting and whitespace changes. The latter is not desirable, as this brings it out of line with other code in this overall project. 

So bits have also been removed and commented out without a clear reason (that I can see). The whole PR is one commit which makes it hard to go through the reasoning in the code. 
Splitting the PR is multiple commits that each have their own piece of work (add mode X, optimize datastructures, handle whitespace etc) would make things easier to review.

fmessmer (changes requested):
marking as "Changes Requested"
see https://github.com/ipa320/cob_driver/pull/411#pullrequestreview-364693335

PR #412, opened 2020-02-21
add topic parameters

Techbeard:
- add Parameters for topic: /scan ; /scan_standy ; /diagnostics


destogl:
@Techbeard: why do you need this? You can just use [the remapping of ROS](https://wiki.ros.org/roslaunch/XML/remap).

Techbeard:
ok nice. I dont know this

PR #414, opened 2020-05-16
Remove leading slashes from frame ids

nlamprian:
Fixes issues resulting from publishing a tf tranformation with frame ids that start with a slash character. Specifically, it solves the problem of cartographer_ros crashing with

> F0516 01:49:25.690616   949 sensor_bridge.cc:33] Check failed: frame_id[0] != '/' ('/' vs. '/') The frame_id /base_footprint should not start with a /. See 1.7 in http://wiki.ros.org/tf2/Migration.


fmessmer (approved):
I guess this is safe to merge
thanks for the contribution :+1:

PR #422, opened 2021-04-12
use default libvlc configuration for noetic

fmessmer:
under ROS Noetic/Ubuntu Focal, the vlc did not open at all and the mimic node instantly exits

the default config is a quick-fix and at least allows to create a libvlc instance - however, the configs are not optimal and need further tuning

@benmaidel do you have any insights into the libvlc difference between the kinetic and the noetic version?


fmessmer:
not perfect...but allows to have green diagnostics for mimic...will optimize once we actually use this on a noetic robot...

PR #423, opened 2021-04-17
ROS_PYTHON_VERSION conditional dependency for `python-tk` and `python-numpy`

mikaelarguedas:
Introspecting Python 2 dependencies in ROS Noetic, found that this package depends on `python-tk` and `python-numpy` instead of its python3 counterpart.

This PR makes the dependency conditional on the ROS_PYTHON_VERSION.
SImilar to https://github.com/ipa320/cob_command_tools/pull/302

This is a semi-automated PR that has not been tested

### DIFF
--- a/cob_bms_driver/package.xml
+++ b/cob_bms_driver/package.xml
@@ -23,7 +23,8 @@
 
   <exec_depend>cob_msgs</exec_depend>
   <exec_depend>cob_srvs</exec_depend>
-  <exec_depend>python-numpy</exec_depend>
+  <exec_depend condition="$ROS_PYTHON_VERSION == 2">python-numpy</exec_depend>
+  <exec_depend condition="$ROS_PYTHON_VERSION == 3">python3-numpy</exec_depend>
   <exec_depend>rospy</exec_depend>
 
 </package>
--- a/cob_bms_driver/package.xml
+++ b/cob_bms_driver/package.xml
@@ -1,5 +1,8 @@
 <?xml version="1.0"?>
-<package format="2">
+<?xml-model
+  href="http://download.ros.org/schema/package_format3.xsd"
+  schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
   <name>cob_bms_driver</name>
   <version>0.7.5</version>
   <description>
@@ -8,10 +11,11 @@
 
   <maintainer email="felixmessmer@gmail.com">Felix Messmer</maintainer>
   <maintainer email="florian.weisshardt@ipa.fraunhofer.de">Florian Weisshardt</maintainer>
-  <author email="mig-mc@ipa.fhg.de">mig-mc</author>
-  <author email="mathias.luedtke@ipa.fraunhofer.de">Mathias Lüdtke</author>
 
   <license>Apache 2.0</license>
+
+  <author email="mig-mc@ipa.fhg.de">mig-mc</author>
+  <author email="mathias.luedtke@ipa.fraunhofer.de">Mathias Lüdtke</author>
 
   <buildtool_depend>catkin</buildtool_depend>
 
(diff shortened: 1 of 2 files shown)


cob_voltage_control/package.xml
```diff
@@ -25,7 +25,8 @@
 
   <exec_depend condition="$ROS_PYTHON_VERSION == 2">python-matplotlib</exec_depend>
   <exec_depend condition="$ROS_PYTHON_VERSION == 3">python3-matplotlib</exec_depend>
-  <exec_depend>python-tk</exec_depend>
+  <exec_depend condition="$ROS_PYTHON_VERSION == 2">python-tk</exec_depend>
+  <exec_depend condition="$ROS_PYTHON_VERSION == 3">python3-tk</exec_depend>
```
fmessmer, line 29: the key `python3-tk` does not exist in rosdistro
[`python-tk`] tries to distinguish between `python-tk` and `python3-tk` but uses `python-tk` with the joker notation `*` and `focal` is missing to be listed

fmessmer (changes requested):

cob_voltage_control/package.xml
mikaelarguedas: It does exist though:
https://github.com/ros/rosdistro/blob/a5f75c7667d603f643e87f72d5d5f8bbbf36dfb1/rosdep/python.yaml#L7327

mikaelarguedas: AFAIU  the use of `<distro>_python3` OSes is deprecated in rosdep because it requires users to override their OS definition with `--os` when invoking rosdep for it to take effect (which is not the case for debs).
The recommended way is to use the `python3-<...>` rosdep key for python3 dependencies et `python-<...>` for the python2 ones and use a condition if a package needs to use one or the other based on an external condition (like the ROS_PYTHON_VERSION)

cob_bms_driver/package.xml
```diff
@@ -23,7 +23,8 @@
 
   <exec_depend>cob_msgs</exec_depend>
   <exec_depend>cob_srvs</exec_depend>
-  <exec_depend>python-numpy</exec_depend>
+  <exec_depend condition="$ROS_PYTHON_VERSION == 2">python-numpy</exec_depend>
+  <exec_depend condition="$ROS_PYTHON_VERSION == 3">python3-numpy</exec_depend>
```
fmessmer, line 31: needs package format 3

cob_voltage_control/package.xml
fmessmer: oh, I missed that

because our travis failed with:
```
install_target_dependencies
0.26sinstall_target_dependencies
'install_target_dependencies' returned with code '1' after 0 min 0 sec
The command ".industrial_ci/travis.sh" exited with 1.
Done. Your build exited with 1.
```
but thats due to the wrong package format in cob_bms_driver

fmessmer (changes requested):

mikaelarguedas:
oh yeah totally missed updating the package format of that one.
updated in https://github.com/ipa320/cob_driver/pull/423/commits/f25a9d75cddd3649cce5145008c53b72cbcd957d

fmessmer (approved):
thanks for iterating @mikaelarguedas

fmessmer:
@mikaelarguedas 
as a side question:
do you need me to trigger a new releases soon? i.e. is the `python-*` (python2) dependency causing any issues on the build farm?
or where did you observe issues that made you propose these changes?

mikaelarguedas:
>  do you need me to trigger a new releases soon? i.e. is the python-* (python2) dependency causing any issues on the build farm?

it's independant from the buildfarm so no rush. I was just wondering why python 2 kept getting installed on my machines, after fixing a couple packages I was using I decided to report or fix all the noetic packages that were installing python2 to avoid having surprises in the future. But I am not currently using any of the cob packages.

PR #432, opened 2022-05-12
power_state_aggregator now also publishes sensor_msgs.BatteryState

fmessmer:
extend `power_state_aggregator` to also publish `sensor_msgs.BatteryState`

ref https://github.com/mojin-robotics/cob4/issues/2373
ref https://github.com/mojin-robotics/cob4/issues/2385
ref https://github.com/mojin-robotics/mojin_robots/pull/476

### DIFF
--- a/cob_bms_driver/src/power_state_aggregator.py
+++ b/cob_bms_driver/src/power_state_aggregator.py
@@ -20,6 +20,7 @@ import numpy
 import warnings
 from std_msgs.msg import Float64
 from cob_msgs.msg import PowerState
+from sensor_msgs.msg import BatteryState
 
 class PowerStateAggregator():
 
@@ -27,6 +28,7 @@ class PowerStateAggregator():
         # get parameters
         self.current_buffer_size = rospy.get_param('~current_buffer_size', 11) # use odd number to prevent 0.0 mean
         self.pub_power_state = rospy.Publisher('power_state', PowerState, queue_size=1)
+        self.pub_battery_state = rospy.Publisher('battery_state', BatteryState, queue_size=1)
         self.voltage = None
         self.current = None
         self.last_currents = []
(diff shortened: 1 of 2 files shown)


cob_bms_driver/src/power_state_aggregator.py
```diff
@@ -121,18 +123,38 @@ def calculate_time_remaining(self):
...
+            ps.time_remaining = time_remaining
             ps.temperature = self.temperature
+
+            bs = BatteryState()
+            bs.header.stamp = self.last_update
+            bs.voltage = self.voltage
+            bs.current = self.current
+            bs.charge = self.remaining_capacity
+            bs.design_capacity = self.full_charge_capacity
+            bs.percentage = relative_remaining_capacity
```
benmaidel: `power_state.relative_remaining_capacity` is in the range 0.0 - 100.0. `battery_state.percentage` has the range 0.0 - 1.0
```suggestion
            bs.percentage = relative_remaining_capacity / 100.0
```

cob_bms_driver/src/power_state_aggregator.py
```diff
@@ -20,13 +20,15 @@
...
 from std_msgs.msg import Float64
 from cob_msgs.msg import PowerState
+from sensor_msgs.msg import BatteryState
 
 class PowerStateAggregator():
 
     def __init__(self):
         # get parameters
-        self.current_buffer_size = rospy.get_param('~current_buffer_size', 11) # use odd number to prevent 0.0 mean
+        self.current_buffer_size = rospy.get_param('~current_buffer_size', 901) # approx. 60sec, topic rate 15Hz, use odd number to prevent 0.0 mean
```
benmaidel, line 29: Not sure if the buffer_size is now to big. I've used a buffer_size of 11 on the sps_bridge as well. Did the old buffer size made Problems?

benmaidel (changes requested):

cob_bms_driver/src/power_state_aggregator.py
fmessmer: we have always been using `901` in the launch file https://github.com/mojin-robotics/mojin_robots/blob/kinetic-devel/mojin_bringup/drivers/bms.launch#L23 - I just now played that back into the actual node

benmaidel (approved):

PR #438, opened 2022-10-24
minimal emulation for phidget_sensors node

fmessmer:


benmaidel (approved):
As a really minimal sim/emu it is ok. For a more complete sim the current state could be stored and published. (analog_sensors, digitial_sensors)

PR #440, opened 2023-05-16
remove hardcoded angle_increment value

benmaidel:
make cob_scan_unifier angle_increment independent by using the info from the input scan


benmaidel:
Tested with two different angle_increment configs. Used Microscan 3 scanner for the test.

PR #443, opened 2023-12-13
fix(cob_scan_unifier): fix handeling of negative angle_increments

mbeutelspacher:
laser scanners that are upside down often have negative angle_increments (and also angle_max < angle_min).

Since we use angle_max=M_PI angle_min=-M_PI in the unified scan here, we can't just copy the angle increment but need to use the absolute value of angle_increments

 - fixes dying of scan_unifier for negative angle_increments


fmessmer (approved):
looks plausible

@mbeutelspacher could you still post a crash report/stack trace of what exactly happens duriing a crash?

mbeutelspacher:
You get a 

```
terminate called after throwing an instance of 'std::length_error'
  what():  vector::_M_fill_insert
```

because you try resizing a vector to a negative length

